PR #5, opened 2020-03-16
ignore description in application-set

### LINKED ISSUE #3: JunOS application-set description parsing
**Describe the bug**
The JunOS parser appears to error out and skip `application-set` blocks which contain a `description` field.

**To Reproduce**
Attempt to parse something like:
```
applications {
  application test-app {
    protocol tcp;
    destination-port 1;
  }
  application-set test-set {
    description "some description";
    application test-app;
  }
}
```
**Expected behavior**
The JunOS parser parses or skips the `description` element in an `application-set` but otherwise parses and adds the data in the application set as it does with other sets.

marshall-sg:
closes #3


cctechwiz:
Looks good to me.

PR #15, opened 2020-04-07
Add CI workflow

cctechwiz:
This workflow creates a docker based on debian:testing-slim, installs Netmeld development dependencies, configures, and builds the Netmeld repo.

It also includes a badge in the README to see the status of the last build.


cctechwiz:
This only runs the CI pipeline on pushes and pull requests to master. Take a look at https://github.com/cctechwiz-forks/netmeld where I did my testing, to see what the actions looks like while running.

marshall-sg:
I added a bunch of questions, mainly to help me understand this concept and its usage/limitations.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,21 @@
...
+  docker:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
cctechwiz, line 18: These needed to be added in the `Secrets` sections of the repository settings. I currently do not have permission to add them myself. You'll need to go to the [Docker Hub Security Settings](https://hub.docker.com/settings/security) and create a new access token to use as `DOCKER_PASSWORD`.

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,163 @@
...
+      run: cmake --build build -j 2 --target netmeld-${{ matrix.core-modules }}
+
+    - name: Test Core Modules
+      run: |
+        cmake --build build -j 2 --target Test.netmeld-${{ matrix.core-modules }}
+        cd build
+        ctest -L ^Test.netmeld-${{ matrix.core-modules }}$
+    
+    - name: Package Core Modules
+      if: matrix.core-modules != 'datalake-core'
```
marshall-sg, line 69: Is this because we don't have a build script for the Datalake module?

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,163 @@
...
+      run: cmake -S . -B build
+
+    - name: Build Modules
+      run: cmake --build build -j 2 --target netmeld-${{ matrix.modules }}
+
+    - name: Test Modules
+      run: |
+        cmake --build build -j 2 --target Test.netmeld-${{ matrix.modules }}
+        cd build
+        ctest -L ^Test.netmeld-${{ matrix.modules }}$
```
marshall-sg, line 106: Just want to double check that this works as expected.  It seems odd the first `$` is evaluated but the final `$` is not.  I guess they don't actually look for `$` but `${{`...

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,163 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Download All Artifacts
+        uses: actions/download-artifact@v2
+      
+      - name: Prepare For Packaging
+        run: |
+          echo ::set-env name=DATETIME::$(date +%Y%m%d.%H%M%S)
+          find . -iname "*.deb" -print | zip --junk-paths artifacts -@
```
marshall-sg, line 135: Does GitHub require the packaging to be a `zip`?

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,163 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: all-debs
+          path: artifacts.zip
+          
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
marshall-sg, line 147: Do we need to create/populate this?

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,163 @@
...
+      
+      - name: Upload Release Asset
+        id: upload_release_asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./artifacts.zip
+          asset_name: all-debs.zip
```
marshall-sg, line 162: Can we call this something else besides `all-debs.zip`?  Maybe something like `netmeld-DTS-debs.zip` where `DTS` is the date-time-stamp for when the package was created.

Dockerfile
```diff
@@ -20,7 +20,13 @@ RUN apt update \
     libpugixml-dev \
     libpcap0.8-dev \
     python \
+    apt-transport-https \
+    ca-certificates \
```
marshall-sg, line 24: Why is this needed?

.github/workflows/docker-build.yml
marshall-sg: You should be able to do this now.

cctechwiz: They have been added.

.github/workflows/github-ci.yml
cctechwiz: yes, those if-checks can be removed once we have those scripts in place. Otherwise the CI workflow fails and doesn't finish all the other jobs.

.github/workflows/github-ci.yml
cctechwiz: This seems to be working as expected, as far as I can tell.

.github/workflows/github-ci.yml
cctechwiz: No, that was just what we were using before on GitLab. It can technically be any of the standard [Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml) as far as I know. Should we change it to something else?

cctechwiz: Yeah, according to the [Release Asset](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) docs it looks like it can be any of those media types.

.github/workflows/github-ci.yml
cctechwiz: No this is a unique key that is automatically generated each workflow run to authenticate the run as coming from our repo. See [here](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token) for more details if you want.

.github/workflows/github-ci.yml
cctechwiz: I changed the `asset_name` to include the `DTS` as requested.

Dockerfile
cctechwiz: The `actions/chekcout@v2` needs those to checkout our repo. The `runs-on` images have it installed, so we need it inside our container as well for it to work. See this [failed build](https://github.com/cctechwiz-forks/netmeld/actions/runs/96588761) for an example of the error I was getting if you want more context.

.github/workflows/github-ci.yml
marshall-sg: Hmmm, then we could publish the debs themselves.  Maybe we will think on that a bit.

.github/workflows/github-ci.yml
marshall-sg: Did you push your changes and update the pull-request?  I'm not seeing the change reflected in the file.

cctechwiz: Pushed the changes. My bad.

PR #26, opened 2020-05-21
Moving command execution functionality into netmeld core

### LINKED ISSUE #21: Add Command Execution Functionality to Core
Not exactly.  Multiple tools have implemented system level command execution logic with slightly different logic, seemingly because the functionality was not available elsewhere.

Add a system level command execution logic capability to the Core module so current modules can leverage a common functionality.

Currently modules are left to implement their own.

**Additional context**
At least present in the Playbook and Datalake modules, but may be elsewhere.

cctechwiz:
Closes #21


cctechwiz:
I think this is good now. There is still some logic for cmd execution in playbook's cmdRunner. I think it's ok to leave that all there since only playbook uses it right now.

PR #30, opened 2020-06-01
Code Organization Restructure Suggestion

mire-all-rashly:
This PR is primarily focused on more clearly splitting the tool suite into functionally related and logical modules.
- Core: Now contains only the common functionality (e.g., as a library) which is used by most all other modules.  This primarily required the removal of Datastore related functionality.
- Datalake, Datastore, Playbook: Now have their associated library self-contained instead of as a separate folder structure (as if it was a separate project).
- Datastore: Now contains the Importers, Graphers, Exporters, Inserters, and any tools specifically designed towards the manipulation and extraction of data within the data store.
- CLW: Added a top level folder for the tool to simplify the process and align with the concept of module then library and/or tools structure.
- All: Reworked the `deb` creation to leverage CPack instead of custom scripts which should aid in keeping both the build and packaging process in sync.  Split the CMake script into (initial) logical chunks.


marshall-sg:
We already have an open issue to review the docs and man page generation.  We will track that work there.

PR #37, opened 2020-07-24
Use starttyping for better special character handling

marshall-sg:
Regarding `nmdb-export-query` and the `ConTeXt` writer.

This PR addresses an occasional issue where some special characters are not handled (i.e., ignored) when wrapped in a `\type{}` block.  Effectively it changes to use `\starttyping` and `\stoptyping` instead.  I simply commented out the regex code instead of removing it to enable easier fall-back if in the future some of those features are needed.


marshall-sg:
I've reviewed this a few different ways now.  It is ready to commit, so doing so.

PR #38, opened 2020-07-24
Several fixes for deb creation and installing

marshall-sg:
This PR fixes:
- `nmdb-initialize` using the correct new path of `/usr/local/etc/netmeld/schemas` for schemas to load
- initialization of the database on deb `postinst configure`
- cpack using the same install prefix as cmake for consistency between the two methods
- `nmdb-graph-network` image files now install correctly again via cpack


datastore/graphers/nmdb-graph-network/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ target_link_libraries(${TGT_TOOL}
 
 nm_install_bin(${TGT_TOOL})
 
+set(image_dir "images")
```
marshall-sg, line 37: Turn to global variable?

datastore/nmdb-initialize/nmdb-initialize.cpp
```diff
@@ -41,7 +41,7 @@ class Tool : public nmdt::AbstractDatastoreTool
 
     const std::string POSTGRES_DB_NAME {"postgres"};
     const std::string MAC_PREFIX_FILE  {"/usr/share/nmap/nmap-mac-prefixes"};
-    const std::string SCHEMA_PATH {nmfm.getConfPath().string() + "/schema"};
+    const std::string SCHEMA_PATH {nmfm.getConfPath().string() + "/schemas"};
```
marshall-sg, line 44: Use the set variable instead of hardcoding.

PR #39, opened 2020-08-03
Show mac updates

### LINKED ISSUE #33: Import tool for `show mac address table` not working for all test cases.
Some Cisco devices appear to flip the column order.  Current tooling expects "vlan, mac, type, port" where sometimes it can be "vlan, mac, port, type" instead.

marshall-sg:
This PR resolves #33 
- Multiple known, documented formats are now supported.
- Unit testing added and covers the various formats.
- Test naming more clearer as it includes tool/library plus file.


marshall-sg:
FYI: Regression testing will flag a difference between current tooling and the new tooling output.  This is expected since it is successfully parsing data now (i.e., the search for too small or `ERROR` output is non-existent for the new tooling now whereas it flags on several files for the old tooling).

PR #42, opened 2020-08-25
nmap parse logic for nbstat and smb-os-discovery

### LINKED ISSUE #41: Import nmap hostname discovery via targeted host scripts
**Is your enhancement request related to a problem? Please describe.**
Enhance the `nmdb-import-nmap` tool to add processing support of the NSE host scripts `nbstat` and `smb-os-discovery`.  

Specifically, extracting details regarding the hostname as associated with the IP.

None.

**Additional context**
None.

marshall-sg:
This PR closes #41.

---

Specifically, it adds parse logic support to the nmap hostscripts:
- nbstat
- smb-os-discovery

### DIFF
--- a/datastore/importers/nmdb-import-nmap/NmapXmlParser.cpp
+++ b/datastore/importers/nmdb-import-nmap/NmapXmlParser.cpp
@@ -1,0 +1,364 @@
+// =============================================================================
+// Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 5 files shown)


datastore/importers/nmdb-import-nmap/NmapXmlParser.cpp
```diff
@@ -0,0 +1,365 @@
...
+    } else if (std::string("smb-os-discovery") == idValue) {
+      std::string reason  {"nmap smb-os-discovery"};
+      std::string line    {nodeScript.attribute("output").value()};
+      std::regex  regex   {"(Computer name|FQDN): (.+?)\n"};
+      std::smatch match;
+      while (std::regex_search(line, match, regex)) {
+        ipAddr.addAlias(std::string(match[2]), reason);
+        line = match.suffix();
+      }
+    }
```
cctechwiz, line 171:
```suggestion
    std::string idValue {nodeScript.attribute("id").value()};
    std::string reason  {"nmap " + idValue};
    std::string line    {nodeScript.attribute("output").value()};
    std::regex  regex;
    std::smatch match;

    if (std::string("nbstat") == idValue) {
      regex = "(NetBIOS name): ([^,]+),";
    } else if (std::string("smb-os-discovery") == idValue) {
      regex = "(Computer name|FQDN): (.+?)\n";
    }

    while (std::regex_search(line, match, regex)) {
      ipAddr.addAlias(std::string(match[2]), reason);
      line = match.suffix();
    }
```

marshall-sg: Ya...it didn't start as a copy/paste.  It actually looked fairly different but I start abstracting things and got them to a "same" point.  I didn't do something closer to your changes after that as I was unsure if we wanted to try to just extract the `element` nodes (for `smb-os-discovery`) instead of doing regex on the `output` string.  The `nbstat` doesn't have those nice nodes, so I was debating one way or the other.  Since the code is already there, let's go with what you got.

PR #45, opened 2020-09-02
Add nmapxmlparser tests

cctechwiz:
Test have been added and debug printing for nmap objects was fixed.

Need some input on testing `extractNseAndSsh`.


cctechwiz:
@marshall-sg I mentioned this in the PR details but thought I should tag it here too.

The `testExtractNseAndSsh` needs to be finished. I spent a decent amount of time on it and it seems really complicated to test well. Just wanted to get your input about this.

cctechwiz:
I was definitely taking too big of a bite out of the nse and ssh tests initially. This is ready for review.

marshall-sg:
I just noticed this, but I think we should rename `NmapXmlParser` to `ParserNmapXml` to be consistent with the other parser naming structure.  I can make the change while I am reviewing it if that sounds fine to you?

cctechwiz:
Sounds good to me.

marshall-sg (approved):
Looks good.

PR #53, opened 2021-01-08
Fix readme files

### LINKED ISSUE #46: Fix README.md files
In general a few of these appear to be outdated or slightly wrong.

Run`man (MODULE|TOOL)_NAME`  to see what they look like after building.

Many of these change revolve around requiring `--device-id` or not.

marshall-sg:
This should resolve #46.

---
There was actually more wrong than what was stated in the issue.
* In various spots, the data was outdated or in the wrong location with regards to the topic it was discussing.
* There were several *new* features that were not documented at all.
* Verbiage was (attempted at least) to be mostly standardized across all the tools.  Mainly this entailed ensuring they all touched on the same topics, calling out exceptions to the norm, etc.
* Examples were updated to be more usable and call out some more interesting use cases where possible.
* Man page generation and the pages themselves have been reworked to point images to the git repo to see them.  Mainly this is so there isn't just a blank space where an image should be.


marshall-sg (commented):
Reviewed and looks good.

Issue #63, opened 2021-01-19
Add support for bare command to `nmdb-import-ipconfig`


marshall-sg:
**Is your enhancement request related to a problem? Please describe.**
The `nmdb-import-ipconfig` tool should be able to process at least some information from a bare `ipconfig` command.  It currently supports when the `/all` or `/allcompartments /all` flags are given, but not when no arguments are given.

**Describe the solution you'd like**
The tool supports the bare `ipconfig` as well as the other formats.

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #78, opened 2021-05-14
Ipconfig bare i63

cmwill:
Related to issue #63  
Adds support for parsing the `ipconfig` command without `/all` or `/all /allcompartments`

### DIFF
--- a/datastore/importers/nmdb-import-ipconfig/Parser.cpp
+++ b/datastore/importers/nmdb-import-ipconfig/Parser.cpp
@@ -73,14 +73,14 @@ Parser::Parser() : Parser::base_type(start)
      | ("Connection-specific DNS Suffix" >> dots > -token
           [(pnx::bind(&Parser::setIfaceDnsSuffix, this, qi::_1))] > qi::eol)
      | ("Default Gateway" >> dots > *(getIp
-          [(pnx::bind(&Parser::addRoute, this, qi::_1))] > qi::eol) > -qi::eol)
+          [(pnx::bind(&Parser::addRoute, this, qi::_1))] > qi::eol))
      | servers
      | ignoredLine
     ) >> *qi::eol
     ;
 
   ifaceTypeName =
-    (token >> "adapter" >>
+    (ifaceType >> "adapter" >>
      qi::as_string[qi::lexeme[+(qi::char_ - ':')]] > ':')
        [(pnx::bind(&Parser::addIface, this, qi::_2, qi::_1))]
     ;
@@ -114,6 +114,10 @@ Parser::Parser() : Parser::base_type(start)
     +qi::ascii::graph
     ;
 
+  ifaceType =
+    +(qi::ascii::print - "adapter")
+    ;
+
   ignoredLine =
     +(qi::char_ - qi::eol) > qi::eol
     ;
@@ -121,6 +125,7 @@ Parser::Parser() : Parser::base_type(start)
   BOOST_SPIRIT_DEBUG_NODES(
       //(start)
       (compartmentHeader)
+      (hostData)
       (adapter)(ifaceTypeName)
       (servers)
       (ipLine)(getIp)
@@ -146,8 +151,13 @@ void
 Parser::addIface(const std::string& _name, const std::string& _type)
 {
   nmdo::Interface iface;
+  std::string whitespace = "\t\n\v\f\r ";
+  std::string type = _type;
+  type = type.erase(type.find_last_not_of(whitespace) + 1);
+  type = type.erase(0, type.find_first_not_of(whitespace));
+  
   iface.setName(_name);
-  iface.setMediaType(_type);
+  iface.setMediaType(type);
   iface.setUp();
 
   curIfaceName = iface.getName();
--- a/datastore/importers/nmdb-import-ipconfig/Parser.cpp
+++ b/datastore/importers/nmdb-import-ipconfig/Parser.cpp
@@ -50,7 +50,6 @@ Parser::Parser() : Parser::base_type(start)
     *(ignoredLine) // Skip garbage after
     ;
 
-
   compartmentHeader =
     +qi::lit('=') > qi::eol > ignoredLine > +qi::lit('=') > qi::eol
     ;
--- a/datastore/importers/nmdb-import-ipconfig/Parser.cpp
+++ b/datastore/importers/nmdb-import-ipconfig/Parser.cpp
@@ -36,20 +36,31 @@ Parser::Parser() : Parser::base_type(start)
     // Skip garbage before
     *(!qi::lit("Windows IP Configuration") >> -qi::omit[+token] >> qi::eol) >>
     (qi::lit("Windows IP Configuration") > +qi::eol >>
-     +(-compartmentHeader >>
-       +( ("Host Name" >> dots > fqdn > qi::eol)
-            [(pnx::bind(&Parser::addDevInfo, this, qi::_1))]
-        | ("Primary Dns Suffix" >> dots > -fqdn > qi::eol)
-        | ignoredLine
-       ) >> qi::eol >>
-       *(adapter)
-      )
-    ) [(qi::_val = pnx::bind(&Parser::getData, this))] >>
+     (
+      +adapter |
+      +(
+        
+        -compartmentHeader >>
+        hostData >>
+        *(adapter)
+      
+       )
+     )
+    ) [qi::_val = pnx::bind(&Parser::getData, this)] >>
     *(ignoredLine) // Skip garbage after
     ;
 
+
   compartmentHeader =
     +qi::lit('=') > qi::eol > ignoredLine > +qi::lit('=') > qi::eol
+    ;
+  
+  hostData = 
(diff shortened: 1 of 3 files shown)


datastore/importers/nmdb-import-ipconfig/Parser.cpp
```diff
@@ -36,21 +36,31 @@ Parser::Parser() : Parser::base_type(start)
...
+      +adapter |
+      +(
+        
+        -compartmentHeader >>
+        hostData >>
+        *(adapter)
+      
+       )
+     )
+    ) [qi::_val = pnx::bind(&Parser::getData, this)] >>
```
marshall-sg, line 49: This logic is wrong.  I sent you two sample files.  While it parses the first, it fails to parse the second config.  Use `-v9` to trace it; it appears to be an issue with the second adapter processing.

datastore/importers/nmdb-import-ipconfig/Parser.cpp
```diff
@@ -36,21 +36,31 @@ Parser::Parser() : Parser::base_type(start)
...
-            [(pnx::bind(&Parser::addDevInfo, this, qi::_1))]
-        | ("Primary Dns Suffix" >> dots > -fqdn > qi::eol)
-        | ignoredLine
-       ) >> qi::eol >>
-       *(adapter)
-      )
-    ) [(qi::_val = pnx::bind(&Parser::getData, this))] >>
+     (
+      +adapter |
+      +(
```
marshall-sg, line 41: Since both of these are at least one, you probably want to pull the `+` to the outside.  I am assuming you mean one or more of either and not one or more of only one section.

marshall-sg (changes requested):
See my comments on the code, but in addition to that:
- There is also a section which you didn't change that is impacted.  The new sample configs don't have a `Host Name` field as typical when of our prior examples.  Thus the logic of the `DNS Suffix` is wrong as it attempts to prepend an empty hostname since the parser normally would have a value for it.
- You probably want to create a unit test for the parser to test the logic.  You can see a sample at `datastore/importers/nmdb-import-ping/Parser.test.cpp`.  I can also create it for you if want, but it will take me a bit to get time for it.

cmwill:
I think that I've fixed it such that it can parse ipconfig output without failing, though I still need to go back and account for the changes in fields when parsing bare ipconfig vs ipconfig /all.

datastore/importers/nmdb-import-ipconfig/Parser.cpp
cmwill: The thought here is that they would be mutually exclusive. With a bare ipconfig, it would only be adapters, while one with added options would then include optional compartments or the host data.

datastore/importers/nmdb-import-ipconfig/Parser.cpp
```diff
@@ -63,14 +73,14 @@ Parser::Parser() : Parser::base_type(start)
      | ("Connection-specific DNS Suffix" >> dots > -token
           [(pnx::bind(&Parser::setIfaceDnsSuffix, this, qi::_1))] > qi::eol)
      | ("Default Gateway" >> dots > *(getIp
-          [(pnx::bind(&Parser::addRoute, this, qi::_1))] > qi::eol) > -qi::eol)
+          [(pnx::bind(&Parser::addRoute, this, qi::_1))] > qi::eol))
```
marshall-sg, line 76: This probably needs to be something like:
```
| ("Default Gateway" >> dots > (+(getIp [(pnx::bind(&Parser::addRoute, this, qi::_1))] > qi::eol) | qi::eol)
```
Primarily because "Default Gateway" can have either 1) one or more IPs or 2) nothing after it (e.g., https://social.technet.microsoft.com/Forums/getfile/828925) and the current logic (in this revision) fails to account for the 2nd case now.

Note: The previous logic would still be wrong given it can be too greedy when "Default Gateway" is the last configuration setting since it will consume two newlines instead of the intended one.

datastore/importers/nmdb-import-ipconfig/Parser.cpp
marshall-sg: Hmmm, that's a fair thought process.  Hopefully they are mutually exclusive.  I believe the reasoning for the `/allcompartments` flag was was abandoned right after it was introduced (i.e., I haven't seen it even documented since Windows 7).  However, who knows what they intend to do since they haven't removed it yet.

marshall-sg (changes requested):

cmwill:
I made some fixes/changes based on your suggestions, though I haven't made any test cases yet as I would need to make the parser rules protected rather than private, and I would rather save that for last. I think this does cover all but the tests.

As for tests, I was thinking of covering these sorts of cases. Do you have any ideas for others that should be covered?

1. Bare
2. /allcompartments /all
3. /all
4. An adapter interface type with multiple words (the Wireless LAN interface in previous test cases caused trouble, and adapters with multiword types didn't seem to be parsed correctly)
5. The last adapter has 0 IPs in its Default Gateway
6. The last adapter has 1 IP in its Default Gateway
7. The last adapter has 2+ IPs in its Default Gateway

datastore/importers/nmdb-import-ipconfig/Parser.cpp
```diff
@@ -136,8 +151,13 @@ void
...
 {
   nmdo::Interface iface;
+  std::string whitespace = "\t\n\v\f\r ";
+  std::string type = _type;
+  type = type.erase(type.find_last_not_of(whitespace) + 1);
+  type = type.erase(0, type.find_first_not_of(whitespace));
+  
   iface.setName(_name);
-  iface.setMediaType(_type);
+  iface.setMediaType(type);
```
marshall-sg, line 160: This works, but I want to ensure you are aware (since I am assuming you are new to Boost Spirit):
If you changed line 118 to `+(qi::ascii::print - (+qi::blank >> "adapter"))` it would pass the interface type without the trailing space(s).  You should never get a space before as `ifaceTypeName`, which calls `ifaceType`, uses a `qi::ascii:blank_type` skipper.

marshall-sg:
> I made some fixes/changes based on your suggestions, though I haven't made any test cases yet as I would need to make the parser rules protected rather than private, and I would rather save that for last. I think this does cover all but the tests.
> 
> As for tests, I was thinking of covering these sorts of cases. Do you have any ideas for others that should be covered?
> 
> 1. Bare
> 2. /allcompartments /all
> 3. /all
> 4. An adapter interface type with multiple words (the Wireless LAN interface in previous test cases caused trouble, and adapters with multiword types didn't seem to be parsed correctly)
> 5. The last adapter has 0 IPs in its Default Gateway
> 6. The last adapter has 1 IP in its Default Gateway
> 7. The last adapter has 2+ IPs in its Default Gateway

I'll prefix: us adding test cases is "new" and we've been doing this as we modify code, so why I made mention of it and why you see there aren't a lot existing.

Generally speaking, what we've been doing is having tests which cover the parts of the parser (this covers probably 4-7 of your list).  So they are more unit test oriented.  Definitely have "good cases" (they pass) and possibly "bad cases" (they fail).  The good cases are basically what the rule is codified to parse.  In some cases, the rules have an attached predicate and you can also examine/test the state of the object after a particular parse path.  If possible, have a few bad cases which cover instance we explicitly are aware of and want to prevent a refactor from allowing.  The parser is pretty good at failing on "bad cases" because of formatting on it's own.

If reasonable, a few interesting whole parser test cases (this covers probably 1-3 of your list) could be added to help a developer understand the general overall structure of the data and some interesting corner cases.  However, in many cases it is not reasonable to codify (e.g., a router/switch config) so it's a judgement call.  We do run changes through a separate, independent, regression testing so it is covered somewhat there as well.

That said, for parts, probably test:
- `compartmentHeader`
  - Good case(s): an arbitrary line of text preceded and followed by a line with only one, or more, equal signs
  - Bad case(s): multiple lines in the center, doesn't start/end with a equal sign line, any other character besides an equal or ending newline in the pre/post-line
- `hostData`
  - Good case(s): tests of host name and dns suffix
  - Bad case(s): ?
- `adapter`
  - Good case(s): your 5-7, but also all the other parts, except `servers` since it has it's own test, of the or block
  - Bad case(s): ?
- `ifaceTypeName`
  - Good case(s): single and multi-word name
  - Bad case(s): ?
- `dots`
  - Good case(s): just a colon, a dot-space colon, and a couple dot-space instances then a colon
  - Bad case(s): ?
- `servers`
  - Good case(s): all the parts of the or block
  - Bad case(s): ?
- `ipLine`
  - Good case(s): three variations on address prefix, test with and without Subnet Mask line afterwards
  - Bad case(s): ?
- `getIp`
  - Good case(s): an IP with zero or one tokens after
    - note: that the IP address parser already tests IP formats so don't worry about that
  - Bad case(s): an IP with multiple tokens after
- `ifaceType`
  - Good case(s): (your 4) single and multi-word type, one or more spaces before adapter
  - Bad case(s): something without adapter after

Don't worry about testing `token` or `ignoredLine` as they are fairly simple and we will ultimately, probably, refactor those to a common Parser logic include for consistency across parsers.

marshall-sg (approved):

Issue #59, opened 2021-01-18
Add config file path option to `nmdb-export-port-list`


marshall-sg:
**Is your enhancement request related to a problem? Please describe.**
It would be good if we can specify the config file path for the `nmdb-export-port-list` tool. This would allow the "default" location to be documented as well as allow an alternate config to be located elsewhere instead of having to alter the default installed one.

**Describe the solution you'd like**
Add the option `--config-path` to the tool.

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #79, opened 2021-05-14
Port list config path i59

cmwill:
#59 

Adds option for specifying an alternate port configuration file in the `nmdb-export-port-list` module

### DIFF
--- a/datastore/exporters/nmdb-export-port-list/nmdb-export-port-list.cpp
+++ b/datastore/exporters/nmdb-export-port-list/nmdb-export-port-list.cpp
@@ -119,9 +119,10 @@ class Tool : public nmdt::AbstractExportTool
             "Use ports from database instead of from resource file.")
           );
           
+      auto& nmfm {nmcu::FileManager::getInstance()};
       opts.addOptionalOption("config-path", std::make_tuple(
             "config-path,c",
-            po::value<std::string>(),
+            po::value<std::string>()->default_value(nmfm.getConfPath()/"port-list.conf"),
             "Use specified port configuration file instead of the default.")
           );
 
@@ -167,7 +168,7 @@ class Tool : public nmdt::AbstractExportTool
       }
       else {
         // Open and parse the port list file.
-        sfs::path const portListPath {this->getConfPath()};
+        sfs::path const portListPath {sfs::path(opts.getValue("config-path"))};
 
         Results portRanges {
           nmdp::fromFilePath<Parser, Results>(portListPath.string())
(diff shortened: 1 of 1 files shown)


datastore/exporters/nmdb-export-port-list/nmdb-export-port-list.cpp
```diff
@@ -118,6 +118,13 @@ class Tool : public nmdt::AbstractExportTool
             NULL_SEMANTIC,
             "Use ports from database instead of from resource file.")
           );
+          
+      opts.addOptionalOption("config-path", std::make_tuple(
+            "config-path,c",
+            po::value<std::string>(),
+            "Use specified port configuration file instead of the default.")
+          );
```
marshall-sg, line 126: Instead of hiding what the default path is, we will want to place it as the default when they display the help messages.  For example,
```suggestion
      auto& nmfm {nmcu::FileManager::getInstance()};
      opts.addOptionalOption("config-path", std::make_tuple(
            "config-path,c",
            po::value<std::string>()->default_value(nmfm.getConfPath()/"port-list.conf"),
            "Use specified port configuration file instead of the default.")
          );
```

This also means the changes below need to be fixed/undone/changed since that logic isn't needed or is invalid now.

datastore/exporters/nmdb-export-port-list/nmdb-export-port-list.cpp
```diff
@@ -160,8 +167,7 @@ class Tool : public nmdt::AbstractExportTool
       }
       else {
         // Open and parse the port list file.
-        auto& nmfm {nmcu::FileManager::getInstance()};
-        sfs::path const portListPath {nmfm.getConfPath()/"port-list.conf"};
+        sfs::path const portListPath {this->getConfPath()};
```
marshall-sg, line 170: If you change the prior, this needs to be fixed to get the value of the configuration option.  For example:
```suggestion
        sfs::path const portListPath {opts.getValue("config-path")};
```

datastore/exporters/nmdb-export-port-list/nmdb-export-port-list.cpp
```diff
@@ -250,6 +256,15 @@ class Tool : public nmdt::AbstractExportTool
...
       return nmcu::Exit::SUCCESS;
     }
+
+    sfs::path getConfPath() {        
+      auto& nmfm {nmcu::FileManager::getInstance()};
+      if (opts.exists("config-path")) {
+        return sfs::path(opts.getValue("config-path"));
+      }
+      return nmfm.getConfPath()/"port-list.conf";
+    }
```
marshall-sg, line 266: If you change the prior two, this can completely go away I believe.
```suggestion
```

marshall-sg (changes requested):
A few changes are requested, but shouldn't be anything major.

cmwill:
I went through and made these changes in the nmdb-export-port-list code. I think that covers the suggested changes unless more come up.

marshall-sg (approved):
All comments resolved.

Issue #72, opened 2021-01-26
Traceroute support with IP linkage graphing


marshall-sg:
**Is your feature request related to a problem? Please describe.**
Traceroute like tools can show the hops along the way, but Netmeld does not support parsing of that data nor do graphing tools support displaying linkage unless subnets are explicitly known.

**Describe the solution you'd like**
Create a tool called `nmdb-import-traceroute` which:
- process a `traceroute` or `traceroute -n` type command (equivalant tracert for Windows)
- IPs would be imported as at least responding hosts and probably as router IPs
- add the next hop type information into the data store

Update the `nmdb-graph-network` to:
- Allow showing linkage between nodes (i.e. next hop) despite a common subnet not being known.

**Describe alternatives you've considered**
Manually adding the data and manually adding connections after the generation by the graphing tool.

**Additional context**
None.

PR #84, opened 2021-09-23
Traceroute i72

cmwill:
#72 

Adds support for parsing output of traceroute [ -n] and tracert [-d] commands and storing the recorded hops

`nmdb-graph-network` inserts directed edges to show hops between devices


marshall-sg (changes requested):
Lots of comments.  Mixture of some logic errors, parsing nuances, style, etc.  Let me know if I need to clarify anything specifically

marshall-sg (approved):
This appears to be good enough.  There are still some improvements that could be made, but none I saw were necessary.

PR #101, opened 2022-01-31
Various bug fixes

mire-all-rashly:
This resolves a few bugs I've encountered during usage that have crept in during recent dev commits as well an issue I was encountering with a tool.
- The `nmdb-import-ip-address` tool did not handle interface entries with `altname` lines; it should be fixed now (see around line 66).  Also added tests for the parser for the tool; nothing fancy, just something to verify I didn't change anything as I was going about it.  There looks like a lot of changes, but it was mainly organizational as I was having a hard time following it and disallowing roll-back because it hid the fact the parser was actually failing to parse correctly.
- Changes to the `nmdb-playbook` allowed for a multi-transaction instantiation instance; should be fixed now since I updated it to close out the transactions after the call (as they are read only anyway).
- Resolved an issue in the `CommandRunner` object where it was testing for non-zero (C++ success) instead of zero (POSIX success) for CLI based execution status.
- The ConTeXt export changes stopped printing lines between inter-network destination IPs.  I think that was a slip as the prior version did and nothing was stated that it was intentional.
- More quality of life/usage than anything, I updated the `plays.yaml` file to turn on debug for namp commands and to have host and script timeouts for slow hosts/scripts.  While the host timeout will kill very slow hosts, the script timeout is the one probably needed more.  Not sure which scripts are doing it, but basically it made my test scans (<20 hosts, ranging 2-30 ports) go from over 1.5 hrs (I manually killed it) to about 38 minutes.

### DIFF
--- a/datastore/importers/nmdb-import-ip-addr-show/Parser.test.cpp
+++ b/datastore/importers/nmdb-import-ip-addr-show/Parser.test.cpp
@@ -115,48 +115,48 @@ BOOST_AUTO_TEST_CASE(testRules)
   }
 
   {
-      const auto& parserRule {tp.iface};
-      // OK
-      std::vector<std::string> testsOk {
-        {R"(1: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
-                link/ether 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff
-                altname enp2s1
-                inet 1.1.1.1/24 scope global ens33
-                   valid_lft forever preferred_lft forever
-                inet6 fe80::1:1111:1:1/64 scope link 
-                   valid_lft forever preferred_lft forever
+    const auto& parserRule {tp.iface};
+    // OK
+    std::vector<std::string> testsOk {
+      {R"(1: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
+              link/ether 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff
+              altname enp2s1
+              inet 1.1.1.1/24 scope global ens33
+                 valid_lft forever preferred_lft forever
+              inet6 fe80::1:1111:1:1/64 scope link 
+                 valid_lft forever preferred_lft forever
 )"},
-        {R"(255: t0: flags mtu 1 tokens
-                link/type 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff opt tokens
-                altname name1
-                altname name2
-                inet 1.1.1.1/24 scope host
-                inet 2.2.2.2/24 scope host
-                inet 3.3.3.3/24 scope host
+      {R"(255: t0: flags mtu 1 tokens
+              link/type 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff opt tokens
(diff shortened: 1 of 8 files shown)


datastore/importers/nmdb-import-ip-addr-show/Parser.test.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <netmeld/datastore/parsers/ParserTestHelper.hpp>
+
+#include "Parser.hpp"
+
+namespace nmdo = netmeld::datastore::objects;
+namespace nmdp = netmeld::datastore::parsers;
+
+using qi::ascii::blank;
+
+class TestParser : public Parser {
```
marshall-sg, line 40:
```suggestion
class TestParser : public Parser
{
```

marshall-sg, line 40:
```suggestion
class TestParser : public Parser
{
```
Style issue; simple.

datastore/importers/nmdb-import-ip-addr-show/Parser.test.cpp
```diff
@@ -0,0 +1,161 @@
...
+                link/type
+                inet 1.1.1.1/24 scope host
+)"},
+      };
+      for (const auto& test : testsOk) {
+        nmdo::Interface out;
+        BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank),
+                   "Parse rule 'inetLine': " << test);
+        BOOST_TEST(out.isValid());
+      }
```
marshall-sg, line 159:
```suggestion
  {
    const auto& parserRule {tp.iface};
    // OK
    std::vector<std::string> testsOk {
      {R"(1: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
              link/ether 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff
              altname enp2s1
              inet 1.1.1.1/24 scope global ens33
                 valid_lft forever preferred_lft forever
              inet6 fe80::1:1111:1:1/64 scope link 
                 valid_lft forever preferred_lft forever
)"},
      {R"(255: t0: flags mtu 1 tokens
              link/type 00:11:22:33:44:55 brd ff:ff:ff:ff:ff:ff opt tokens
              altname name1
              altname name2
              inet 1.1.1.1/24 scope host
              inet 2.2.2.2/24 scope host
              inet 3.3.3.3/24 scope host
)"},
      {R"(255: t0: flags mtu 1
              link/type
)"},
      {R"(255: t0: flags mtu 1
              link/type brd ff:ff:ff:ff:ff:ff
)"},
      {R"(255: t0: flags mtu 1
              link/type brd ff:ff:ff:ff:ff:ff opt tokens
)"},
      {R"(255: t0: flags mtu 1
              link/type
              altname name1
)"},
      {R"(255: t0: flags mtu 1
              link/type
              inet 1.1.1.1/24 scope host
)"},
    };
    for (const auto& test : testsOk) {
      nmdo::Interface out;
      BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank),
                 "Parse rule 'inetLine': " << test);
      BOOST_TEST(out.isValid());
    }
```
Style issue; simple.

marshall-sg (changes requested):
Only style changes wanted.

marshall-sg (approved):
Looks good.

PR #102, opened 2022-02-01
Post tool chain logic

### LINKED ISSUE #88: Post-processing Tool Chain Logic
It would be nice to have a post-processing/importing tool(s) which could pull out items of interest/concern.  Similar to how the graphing tools export graphic views of the network connectivity, something which exports textual/graphic views of data that may be of interest.

Effectively, a tool(s) which can take an input file(s) of things to gather from the data-store/lake and spit out a report(s).  While it could be built into the tool, it probably would be more useful to allow the end-user to define the queries.

The input file (user generated) probably should contain:
- pre-data intro (e.g., this is what/significance of what follows)
- query to run
- (stretch goal'ish) possible sub area/queries

The output file (tool generated) probably should contain:
- sectioning based on the queries ran
- query output
- (stretch goal'ish) be in ConTeXt(?) and standalone build-able

Manually running similar query/logic chains.

**Additional context**
None.

marshall-sg:
Resolves #88 and then some.
---
Specifically for issue:
- Adds a tool, currently named `nmdb-analyze-data`, to allow for repeatable command execution.  While this probably could easily be done with a script as well, this does allow a somewhat easier tie-in to the tooling back-end.
- Included sample file with several initial handy queries for general analysis.

Other:
- The dep `libyaml-cpp` got a version bump, tooling requires nothing different but the old is version is not in the repos anymore so need to update.
- Added a couple of new functions for getting Datastore related information and updated associated tooling to leverage where possible.
- Updated `nmdb-import-ip-addr-show` tool to not require an EOL before EOF.
- Resolved some verbiage inconsistencies and logic leftovers (i.e., previously open issues which were resolvable give regression corpus) identified by the `ToolObservations`.
- Updated Playbook plays file to correctly include both port scan and device config discovered possible DNS servers.


marshall-sg (commented):
Looks ok at this point.

marshall-sg (commented):
Looks good.

PR #103, opened 2022-02-04
94-cppcheck

### LINKED ISSUE #94: Resolve cppcheck Issues
**Describe the bug**
Numerous potential issues where introduced with the new route logic support.  While not perfect (as it doesn't support code after c++11) and all of them may not be actual issues, run the `cppcheck` to see the issue and mitigate as many as possible.

**To Reproduce**
Effectively, build the code via
 `cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON && cmake --build build -j4 --target Test.netmeld && (cd build && ctest Test.netmeld)` and then run the first checks via
`cppcheck --template=gcc --enable=all --force --project=build/compile_commands.json --inline-suppr -Dfalse=0 --suppressions-list=cppcheck-supressions.txt . > /dev/null`
and (to ensure the suppressed checks/list still isn't a real issue) then
`cppcheck --template=gcc --enable=all --force --project=build/compile_commands.json --inline-suppr -Dfalse=0 . > /dev/null`.

**Expected behavior**
Output should only be those that can't be mitigated.

**Screenshots**
None.

**Additional context**
None.

cmwill:
Resolves #94


cmwill:
Cppcheck does not seem to account for one variable (usually the key) being unused when a map is being iterated using newer syntax.

Some cases of unused functions that cppcheck noted could possibly be used in the future, was unsure of whether they should be removed.

Inline suppressions may possibly require executing cppcheck with the `--inline-suppr` argument.

marshall-sg:
Maybe these were just notes/comments to yourself, but:

> Cppcheck does not seem to account for one variable (usually the key) being unused when a map is being iterated using newer syntax.

Yes, that's what I was seeing as well.  Keep things like this in mind as `cppcheck` hasn't caught up to the latest standard so it may suggest things differently when it is perfectly fine due to improvements on the standard (since C++11).  We should generally ignore the issues where it is valid syntax in the standard we use (C++20).

> Some cases of unused functions that cppcheck noted could possibly be used in the future, was unsure of whether they should be removed.

We probably want to get rid of them, but I would be interested in knowing what/where they are just to double check.

> Inline suppressions may possibly require executing cppcheck with the `--inline-suppr` argument.

Expected and part of our notes for using `cppcheck` (see the issue and the `cppcheck-supressions.txt` file).  Basically because of the standard lag and more or less non-support/understanding of Boost Spirit, we have to use suppression to quite things it flags but we have verified as being false-positives.

cmwill:
They weren't just to myself, mainly code flagged by cppcheck that I wasn't going to handle until I got some clarification.

A mutex lock was flagged but I haven't done anything with it yet, unsure of if it's guarding against multiple instances of the singleton being used or not.

```
datastore/importers/nmdb-import-tshark/DataContainerSingleton.cpp:43:31: warning: The lock is ineffective because the mutex is locked at the same scope as the mutex itself. [localMutex]
  std::lock_guard<std::mutex> lock(tmp);
```

These are the unused functions flagged.

```
core/common/utils/ProgramOptions.cpp:86:0: warning: The function 'addConfFileOption' is never used. [unusedFunction]

^
datastore/common/objects/AclPortSet.cpp:57:0: warning: The function 'addPortRange' is never used. [unusedFunction]

^
core/common/utils/ProgramOptions.cpp:120:0: warning: The function 'setConfFile' is never used. [unusedFunction]

^
datastore/common/objects/Service.cpp:66:0: warning: The function 'setDstFqdn' is never used. [unusedFunction]

^
datastore/common/objects/DnsResolver.cpp:70:0: warning: The function 'setDstPort' is never used. [unusedFunction]

^
datastore/common/objects/AclRulePort.cpp:57:0: warning: The function 'setDstPortSetId' is never used. [unusedFunction]

^
datastore/common/objects/DnsResolver.cpp:46:0: warning: The function 'setIface' is never used. [unusedFunction]

^
datastore/common/objects/DnsResolver.cpp:52:0: warning: The function 'setScopeDomain' is never used. [unusedFunction]

^
datastore/common/objects/AclRulePort.cpp:51:0: warning: The function 'setSrcPortSetId' is never used. [unusedFunction]

```

marshall-sg:
Roger that.


> A mutex lock was flagged but I haven't done anything with it yet, unsure of if it's guarding against multiple instances of the singleton being used or not.
> 
> ```
> datastore/importers/nmdb-import-tshark/DataContainerSingleton.cpp:43:31: warning: The lock is ineffective because the mutex is locked at the same scope as the mutex itself. [localMutex]
>   std::lock_guard<std::mutex> lock(tmp);
> ```
You are on the right track.  Effectively, it is there to guard against the race condition of multiple calls to the singleton instantiation.  I could be wrong, but I do not believe `cppcheck` is correct in this case since the function and logic in question involve static data.  Technically we could suppress the flagging by placing that into another scope level (i.e., wrap the code in parenthesis), but it doesn't change the logic that I'm aware of.


> These are the unused functions flagged.
> 
> ```
> core/common/utils/ProgramOptions.cpp:86:0: warning: The function 'addConfFileOption' is never used. [unusedFunction]
> datastore/common/objects/AclPortSet.cpp:57:0: warning: The function 'addPortRange' is never used. [unusedFunction]
> core/common/utils/ProgramOptions.cpp:120:0: warning: The function 'setConfFile' is never used. [unusedFunction]
> datastore/common/objects/Service.cpp:66:0: warning: The function 'setDstFqdn' is never used. [unusedFunction]
> datastore/common/objects/DnsResolver.cpp:70:0: warning: The function 'setDstPort' is never used. [unusedFunction]
> datastore/common/objects/AclRulePort.cpp:57:0: warning: The function 'setDstPortSetId' is never used. [unusedFunction]
> datastore/common/objects/DnsResolver.cpp:46:0: warning: The function 'setIface' is never used. [unusedFunction]
> datastore/common/objects/DnsResolver.cpp:52:0: warning: The function 'setScopeDomain' is never used. [unusedFunction]
> datastore/common/objects/AclRulePort.cpp:51:0: warning: The function 'setSrcPortSetId' is never used. [unusedFunction]
> ```
(I snipped the quote down to the important parts)
I went through several of the various code bases we have.  I don't see anyone using them currently.  However for at least some of those functions, I know the intended usage and I'm fairly confident it is "unused" due to the lack of collapsing diverging implementations (i.e., alternate logic was introduced at one point but everything hasn't been migrated yet).  For some of them, the correct path hasn't been determined yet.  So, for the time being just comment the unused functions out to silence the warnings.  You may not be able to comment out the associated objects/logic (acted on by the function but defined elsewhere) as those are required for other logic to be correct.

marshall-sg (approved):
Looks good and all testing passed.

Issue #96, opened 2021-11-29
Docker Builds for Tooling


marshall-sg:
**Is your enhancement request related to a problem? Please describe.**
Add more Docker builds to support easier usage, dev, etc. of tooling and code via docker images.

**Describe the solution you'd like**
Docker image build(s) (integrated into the docker build pipeline) for:
- Entire Netmeld tool suite in one image
- Netmeld modules in images (one image per module)
  - This might not actually make sense for all modules/sub-modules/tools and will need to be examined more
- Some rudimentary testing for each to ensure they can be used (i.e., runtime testing)
  - As best as possible
  - This is a replacement for #77 concepts

**Describe alternatives you've considered**
Manually creating docker images each time.

**Additional context**
None.

PR #110, opened 2022-04-29
More Docker builds for Netmeld

cmwill:
Draft for Issue #96 
Currently untested, but I wanted to make sure the ideas are sound before I continue.

My thoughts were one image with the entire toolsuite, one image with the tools for fetching, and one image with the datalake and datastore, though the third one doesn't have psql setup yet. The playbook module is also only in the full toolsuite image and not one of its own.

I'm not yet sure on how the testing workflows could be handled. I've thought of building an image specifically for testing, but I'm unsure of it that would be enough.

### DIFF
--- a/Dockerfile-datalake-datastore
+++ b/Dockerfile-datalake-datastore
@@ -1,0 +1,140 @@
+FROM debian:testing-slim
+
+RUN apt update \
+ && apt install --assume-yes --no-install-recommends \
+      debconf \
+      build-essential \
+      cmake \
+      make \
+      gcc \
+      g++ \
+      git \
+      help2man \
+      pandoc \
+      libboost-date-time-dev \
+      libboost-iostreams-dev \
+      libboost-program-options-dev \
+      libboost-system-dev \
+      libboost-test-dev \
+      libpqxx-dev \
+      libpugixml-dev \
+      libpcap0.8-dev \
+      nlohmann-json3-dev \
+      libyaml-cpp-dev \
+ && apt install --assume-yes --no-install-recommends \
+      python3 \
+      apt-transport-https \
+      ca-certificates \
+ && rm -rf /var/lib/apt/lists/* \
+ && update-ca-certificates
+
+RUN groupadd -r netmeld \
+ && useradd -r -s /bin/false -g netmeld netmeld
+USER netmeld
+
+WORKDIR /home/netmeld
+
+RUN git clone https://github.com/netmeld/netmeld.git
+
+ENV cmakereplacement '# =============================================================================\n\
+# Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC\n\
+# (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.\n\
+# Government retains certain rights in this software.\n\
+#\n\
+# Permission is hereby granted, free of charge, to any person obtaining a copy\n\
+# of this software and associated documentation files (the "Software"), to deal\n\
+# in the Software without restriction, including without limitation the rights\n\
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell\n\
+# copies of the Software, and to permit persons to whom the Software is\n\
+# furnished to do so, subject to the following conditions:\n\
+#\n\
+# The above copyright notice and this permission notice shall be included in\n\
+# all copies or substantial portions of the Software.\n\
+#\n\
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\n\
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\n\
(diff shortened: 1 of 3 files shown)


docker-testing/Dockerfile-fetchers
```diff
@@ -0,0 +1,17 @@
+FROM netmeld-test-base
+
+ENV NETMELD_DB netmeld_test_fetchers
+
+WORKDIR /home/netmeld/setup
+
+RUN apt install --quiet --quiet --assume-yes --option Dpkg::Use-Pty=0 \
+      --no-install-recommends --no-install-suggests \
+      ./netmeld-*core.deb ./netmeld-*fetchers.deb ./netmeld-*tool-clw.deb
```
marshall-sg, line 9: Couple of things with this,

- While the `--no-install-suggests` might be ok, we want to install recommends. The use cases for not installing them are special corner cases we don't need to test under this Issue/PR.
- We don't want to explicitly call out dependencies that should already be installed prior to building this container.
- There should be a different Dockerfile for each module.

```suggestion
RUN apt install --quiet --quiet --assume-yes --option Dpkg::Use-Pty=0 \
      --no-install-suggests \
      ./netmeld-*fetchers.deb
```

docker-testing/Dockerfile-fetchers
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR /home/netmeld
+
+RUN rm -rf ./setup
+
+WORKDIR /home/netmeld/contents/fetchers
+
```
marshall-sg, line 16: Not needed as this shouldn't be the testing version.
```suggestion
```

docker-testing/Dockerfile-toolsuite
```diff
@@ -0,0 +1,17 @@
...
+RUN apt install --quiet --quiet --assume-yes --option Dpkg::Use-Pty=0 \
+      --no-install-recommends --no-install-suggests \
+      ./netmeld-*.deb
+
+WORKDIR /home/netmeld
+
+RUN rm -rf ./setup
+
+WORKDIR /home/netmeld/contents/
+
```
marshall-sg, line 16: Not needed as this shouldn't be the testing version.
```suggestion
```

docker-testing/Dockerfile-fetchers
```diff
@@ -0,0 +1,17 @@
...
+CMD ["./test.py"]
```
marshall-sg, line 17: Since we aren't using for testing, it should drop to a shell or execute a command.  Furthermore, we should confine it to the `netmeld` user.  So something like:
```suggestion
USER netmeld

CMD ["/bin/bash"]
```

docker-testing/Dockerfile-toolsuite
```diff
@@ -0,0 +1,17 @@
...
+CMD ["./toolsuite.sh"]
```
marshall-sg, line 17: Since we aren't using for testing, it should drop to a shell or execute a command.  Furthermore, we should confine it to the `netmeld` user.  So something like:
```suggestion
USER netmeld

CMD ["/bin/bash"]
```

docker-testing/Dockerfile-fetchers
```diff
@@ -0,0 +1,17 @@
+FROM netmeld-test-base
+
+ENV NETMELD_DB netmeld_test_fetchers
+
```
marshall-sg, line 4: Not seeing why this is here.
```suggestion
```

docker-testing/Dockerfile-toolsuite
```diff
@@ -0,0 +1,17 @@
+FROM netmeld-test-base
+
+ENV NETMELD_DB netmeld_test_all
+
```
marshall-sg, line 4: Not seeing why this is here.
```suggestion
```

docker-testing/contents/datalake-datastore/test.py
```diff
@@ -0,0 +1,444 @@
...
+import subprocess
+import sys
+
+sys.path.append("..")
+from  run_command import run_command
+
+def main():
+
+    db_args = f'--db_args="host={os.environ["NETMELD_DB_HOST"]} user={os.environ["NETMELD_DB_USER"]} password={os.environ["NETMELD_DB_PASS"]}"'
+    db_name = f'--db_name="{os.environ["NETMELD_DB"]}"'
```
marshall-sg, line 13: I didn't see anything in particular which you are doing that requires this to be defined.  The tools provide their own default.  For this Issue/PR, I don't think we need to support supplying custom parameters.
NOTE: This will require removal of these arguments from all following commands.
```suggestion
```

marshall-sg (changes requested):
Several changes have been requested.  On top of those are the following comments:
- Some of the objects outlined in the "Describe the solution you'd like" section were missed.  I've made some comments directly in the code towards this effect.  However just to ensure, the intent is to have Docker images for each module, where it makes sense.  The testing part is an additional Docker image.  Based on what I see, it should be fairly straightforward to refactor the current code to achieve this and it may even simplify the testing logic since it only has to be done in one container versus every container.
-  Despite the ask, don't worry about integrating this into the build pipeline.  We will deal with that issue separately as we also need to be able to push them to the Docker registry for Netmeld.

PR #114, opened 2022-09-20
Aws processing

marshall-sg:
Added initial support for processing AWS output.


marshall-sg:
2x reviewed; looks good.

PR #116, opened 2022-10-20
Update hosts parser

ben-anthony-snl:
The /etc/hosts parser was using an old parser style. This pull request should update it to the modern parser


datastore/importers/nmdb-import-hosts/Parser.cpp
```diff
@@ -0,0 +1,69 @@
...
+Parser::Parser() : Parser::base_type(start)
+{
+  start =
+    *(line)
+    ;
+
+  // (IP address [save to _val] >> domain name [add an alias for the IP in _val that is the result of the parsed ParserDomainName with the reason "hosts file"] >> not a comment >> end of line) | (comment >> end of line) | (end of line)
+  line = (ipAddr [(qi::_val = qi::_1)] >> +domainName[(pnx::bind(&nmdo::IpAddress::addAlias, &qi::_val, qi::_1, "hosts file"))] >> -comment >> qi::eol) | (comment >> qi::eol) | (qi::eol);
+
+  comment = (qi::lit('#') >> *(qi::ascii::graph | qi::ascii::blank));
```
marshall-sg, line 57: Similar to prior rollback comment.

```suggestion
  comment = (qi::lit('#') > *(qi::ascii::graph | qi::ascii::blank));
```

datastore/importers/nmdb-import-hosts/Parser.test.cpp
```diff
@@ -0,0 +1,98 @@
...
+      "127.0.0.1	localhost\n",
+      "127.0.1.1	MyComputer\n",
+      "\n",
+      "# The following lines are desirable for IPv6 capable hosts\n",
+      "::1     ip6-localhost ip6-loopback\n",
+      "fe00::0 ip6-localnet\n",
+      "ff00::0 ip6-mcastprefix\n",
+      "ff02::1 ip6-allnodes\n",
+      "ff02::2 ip6-allrouters\n"
+    };
```
marshall-sg, line 64: There seems to be several similar-type tests here (i.e., `ip hostname` or `comment`).  While not bad, it may be hiding the fact that cases handled by the `line` logic are missing.  More specifically, this test logic should cover the following:
- `ip hostname`
- `ip hostname alias1 alias2 alias3`; really it can have arbitrary alias, but we do not need to explicitly test the arbitrary aspect
- `ip hostname comment`
- `ip hostname alias1 comment`; again arbitrary alias counts
- `comment`
- an empty line

datastore/importers/nmdb-import-hosts/Parser.test.cpp
```diff
@@ -0,0 +1,98 @@
...
+       "192.168.1.10    foo.mydomain.org       foo\n",
+       "192.168.1.13    bar.mydomain.org       bar\n",
+       "146.82.138.7    master.debian.org      master\n",
+       "209.237.226.90  www.opensource.org\n",
+        "\n",
+       "# The following lines are desirable for IPv6 capable hosts\n",
+       "::1             localhost ip6-localhost ip6-loopback\n",
+       "ff02::1         ip6-allnodes\n",
+       "ff02::2         ip6-allrouters\n"
+  };
```
marshall-sg, line 93: This should not be several individual lines, but a single "file" type test.  I've done it with raw string literal (i.e., `R"()"`, https://en.cppreference.com/w/cpp/language/string_literal) type constructs (e.g., see `nmdb-import-ip-addr-show/Parser.test.cpp`), but there are other options as well (e.g., I believe you only need to remove the comma after every line).

marshall-sg (changes requested):
Some small changes.

datastore/importers/nmdb-import-hosts/Parser.cpp
```diff
@@ -0,0 +1,69 @@
...
+// Parser logic
+// =============================================================================
+Parser::Parser() : Parser::base_type(start)
+{
+  start =
+    *(line)
+    ;
+
+  // (IP address [save to _val] >> domain name [add an alias for the IP in _val that is the result of the parsed ParserDomainName with the reason "hosts file"] >> not a comment >> end of line) | (comment >> end of line) | (end of line)
+  line = (ipAddr [(qi::_val = qi::_1)] >> +domainName[(pnx::bind(&nmdo::IpAddress::addAlias, &qi::_val, qi::_1, "hosts file"))] >> -comment >> qi::eol) | (comment >> qi::eol) | (qi::eol);
```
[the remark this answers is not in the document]
ben-anthony-snl: I tried extracting the qi::eol to the start line: `start = *(line > qi::eol)` like you suggested, but that seems to make it fail on a line that is only a new line. As best as I can tell, this is because line can't match the ipAddr part or the comment. Is there a way to mark if nothing matches to just move on to the qi::eol?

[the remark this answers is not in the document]
ben-anthony-snl: Hmmm... Looking through it more, shouldn't it not matter? The fact that we're using the kleene parser on that start line means it should match 0+ times, right? So if it fails the line parse, it should be fine? I'm confused as to how this is failing. Maybe I'm missing something?

marshall-sg (approved):

marshall-sg (approved):

PR #121, opened 2022-12-20
Added rpm parsing tool. Modified Package Object for optional package state

ljvalen:


datastore/importers/nmdb-import-rpm/Parser.cpp
```diff
@@ -0,0 +1,101 @@
...
+    +token
+  ;
+
+  description =
+    +qi::ascii::print
+  ;
+
+  token =
+    +qi::ascii::graph
+  ;
```
marshall-sg, line 67: This isn't intuitive.

Ignoring the rule definition for a moment, by implementation the `token` matches one or more non-space characters, and `packageName`/`version`/`architecture`/`description` matches one or more `token`.  So, for example, the `packageName` rule should technically consume a whole line.  The only reason it doesn't is the rule definition which doesn't define a skipper.  So reading the rule implementation is counter intuitive on what it actually does.

Technically, there are a couple of ways to resolve this but the simplest is probably to remove the `+` from `token` in each of the rules.  Looking back, I think I missed this nugget in the `dpkg` version; we should probably fix that there as well.

datastore/importers/nmdb-import-rpm/Parser.test.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+BOOST_AUTO_TEST_CASE(testWhole)
+{
+  TestParser tp;
+  const auto &parserRule {tp};
+
+  std::vector<std::string> testsOk {
+    // dpkg -l | head
```
marshall-sg, line 73: Isn't this for `rpm -q`?

datastore/importers/nmdb-import-rpm/Parser.test.cpp
```diff
@@ -0,0 +1,91 @@
...
+      "pkg-other  1.1.99.0-5  amd64   more des-cript -- ion\n",
+      "pkg-o12    1.0+v1      x86_64  description\n",
+    };
+
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                 "Parse rule 'packageLine': " << test);
+    }
+  }
+}
```
marshall-sg, line 65: This and the `testWhole` are effectively the same.  If you are going to test parts, then it would be best to check the changed state of the object being populated after each individual rule is called.  Otherwise, just remove this and let the test whole suffice.

datastore/importers/nmdb-import-rpm/README.md
```diff
@@ -0,0 +1,38 @@
+DESCRIPTION
+===========
+
+Parse and import the output of the `rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n"` command on modern Linux
+systems into the Netmeld framework for later analysis. Import's package's
+based on device-id.
```
marshall-sg, line 6:
```suggestion
Parse and import the output of the
`rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n"`
command on modern Linux systems into Netmeld for later analysis.
The imported package list is associated with the passed `--device-id` value.
```

datastore/importers/nmdb-import-rpm/nmdb-import-rpm.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  // ===========================================================================
+  // Constructors
+  // ===========================================================================
+  private:
+  protected:
+  public:
+    Tool() : nmdt::AbstractImportTool<P,R>
+      (
+       "dpkg -l",  // command line tool imports data from
```
marshall-sg, line 53: This is the wrong command.

marshall-sg (changes requested):
See change requests.

datastore/importers/nmdb-import-rpm-query/nmdb-import-rpm-query.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  // ===========================================================================
+  // Constructors
+  // ===========================================================================
+  private:
+  protected:
+  public:
+    Tool() : nmdt::AbstractImportTool<P,R>
+      (
+       "rpm -qa",  // command line tool imports data from
```
marshall-sg, line 53: The import tool won't work unless a specific command is given for a specific output format.  The CLI line needs to be more explicit in that sense.  Something like:
```suggestion
       R"(rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n")",  // command line tool imports data from
```

marshall-sg (changes requested):
Couple minor things, otherwise it looks good.

datastore/importers/nmdb-import-rpm-query/README.md
```diff
@@ -0,0 +1,40 @@
+DESCRIPTION
+===========
+
+Parse and import the output of the 
+`rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n"`
```
marshall-sg, line 5:
```suggestion
`rpm -qa --queryformat="%-50{NAME}%10{VERSION}-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n"`
```

datastore/importers/nmdb-import-rpm-query/README.md
```diff
@@ -0,0 +1,40 @@
...
+prior to processing.
+This could be accomplished manually, by using a tool (e.g., `uni2ascii`),
+or similar.
+
+EXAMPLES
+========
+
+Gather package output from target with
+```
+rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n" > packages.txt
```
marshall-sg, line 22:
```suggestion
rpm -qa --queryformat="%-50{NAME}%10{VERSION}-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n" > packages.txt
```

datastore/importers/nmdb-import-rpm-query/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+or to strip unicode
+```
+rpm -qa --queryformat="%-50{NAME}%10{VERSION}%-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n" | uni2ascii > packages.txt
```
marshall-sg, line 26:
```suggestion
rpm -qa --queryformat="%-50{NAME}%10{VERSION}-%-20{RELEASE}%-20{ARCH}%{SUMMARY}\n" | uni2ascii > packages.txt
```

datastore/importers/nmdb-import-rpm-query/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Process the target data for the device workstation from a local file
+```
+nmdb-import-rpm --device-id <workstation> packages.txt
```
marshall-sg, line 31:
```suggestion
nmdb-import-rpm-query --device-id <workstation> packages.txt
```

datastore/importers/nmdb-import-rpm-query/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Assuming `...` is some command chain which retrieves the target data from a
+remote host and displays the results locally, then the following would
+process it and save the data to a file called `packages.txt` in the current
+working directory.
+```
+... | nmdb-import-rpm --pipe --device-id <workstation> packages.txt
```
marshall-sg, line 39:
```suggestion
... | nmdb-import-rpm-query --pipe --device-id <workstation> packages.txt
```

marshall-sg (changes requested):
I missed a couple of issues in the docs.  I didn't catch the query was changed, the one that was listed won't work as there is an extra `%`.

marshall-sg (approved):
Looks good.

Issue #126, opened 2023-03-13
Datastore object(s) lack appropriate debug output


marshall-sg:
**Describe the bug**
At the least, the `Vrf` object in the Datastore lacks an appropriate debug output.  It currently only displays a `vrfId` instead of all the contents of the object which may be needed for debugging code logic.

There could be other impacted objects (did not do a full inspection).  This issue shouldn't be resolved until any/all are fixed.

**To Reproduce**
Examine/Call the `Vrf::toDebugString()` method; it only returns one variable instead of all the object's variables.

**Expected behavior**
All Datastore module objects should consistently return a string containing all variable identifiers and values for the object of the form:
`[variable1: value, variable2: value]`

**Screenshots**
N/A

**Additional context**
It may be worthwhile to return a debug string as JSON.  However, that may be overkill and only a simple string is really needed.

PR #135, opened 2023-04-27
Vrf debug string now includes interfaces and routes

cmwill:
#126 

Will check other debug strings before marking as ready

### DIFF
--- a/core/common/utils/StreamUtilities.hpp
+++ b/core/common/utils/StreamUtilities.hpp
@@ -32,9 +32,25 @@
 #include <ostream>
 #include <set>
 #include <vector>
+#include <tuple>
 
 
 namespace netmeld::core::utils {
+
+  template<size_t Idx = 0, typename... Types>
+  std::ostream& operator<<(std::ostream& os, const std::tuple<Types...>& tup)
+  {
+    if constexpr(Idx == 0) {
+      os << "[";
+    }
+    os << std::get<Idx>(tup);
+    if constexpr(Idx + 1 < sizeof...(Types)) {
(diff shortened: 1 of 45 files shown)


marshall-sg:
I just looked at some of the purposed changes.  Some of the logic is recreating the wheel so to speak.  Please see the [Core module's](https://github.com/netmeld/netmeld/tree/master/core/common/utils) `StreamUtilities`, as much of the logic for expansion of several container types is already handled.  So just using the variable in an output stream will result in calling the logic.  If you encounter an instance where the container type is not handled, you probably should add it to `StreamUtilities` to ensure consistent handling across all objects rather than one-offs in an object.

core/common/utils/StreamUtilities.hpp
```diff
@@ -32,10 +32,26 @@
...
 
+  template<size_t Idx = 0, typename... Types>
+  std::ostream& operator<<(std::ostream& os, const std::tuple<Types...>& tup)
+  {
+    if constexpr(Idx == 0) {
+      os << "[";
+    }
+    os << std::get<Idx>(tup);
+    if constexpr(Idx + 1 < sizeof...(Types)) {
+      os << " -- ";
```
marshall-sg, line 48: Not sure what the intent was here.  If it was to provide a separator between the tuple values, then please just use a comma then space (`, `) instead of a space, double dash, and space(` -- `).

marshall-sg (changes requested):
Mostly minor changes requested.

marshall-sg (approved):
Looks good, thanks!

Issue #130, opened 2023-03-13
Juniper xml config parser


marshall-sg:
**Describe the bug**
The Datastore's Juniper XML config parser seems to not handle routes which reference other routing tables correctly.  More specifically, if the global `<routing-options>` contains a route who's next-hop is `<next-table>Something.inet.0</next-table>` than the tool will fail to correctly parse and save the route.  

**To Reproduce**
Parse a config with a table defined route, similar to below
```
<routing-options>
  <static>
    <route>
      <name>Table-Defined-Route</name>
      <next-table>Some-Table.inet.0</next-table>
    </route>
  </static>
</routing-options>
```

**Expected behavior**
All routes are parsed and stored in the datastore.

**Screenshots**
N/A

**Additional context**
~It appears that the config is fully parsed and saved, however the route in question seems to be dropped as invalid.~  It appears either the parser is not correctly resolving the route for insertion into the datastore or the datastore needs to be updated to allow for routes referencing other route tables.

PR #136, opened 2023-05-11
Updated DB route, DB ACL, and associated tooling logic

### LINKED ISSUE #131: DB route and acl function logic woes
**Describe the bug**
The functions which determine routability (`ip_route_paths`) and accessibility (`ip_route_path_acls`) appear to be broken and/or working oddly.  They will return routes which are invalid in that they yield routes/paths which don't end with the destination subnet (e.g., default routes are shown to networks not aligned/ending with the subnet).  If it is to be implied that it actually does end in the destination subnet, that is not apparent given where the "hop" ends and what is connected on the other side of the interface (when known in the DB).

Furthermore, the speed at which these functions run is generally unusable for even small networks (order of hours).  It becomes significantly quicker to manually query and cross check than to leverage the functions.  Making any value of the functions lost.

**To Reproduce**
Just running the functions when data is loaded produces the unexpected results.

**Expected behavior**
1. Routing paths are shown with every hope until the final destination subnet is reached and displayed.
    - There most likely should be a differentiation between routing paths which explicitly arrive at the destination subnet (i.e., as all configurations have been imported along the path) versus routing paths which theoretically arrive at the destination (i.e., a default route path is taken and the next-hop subnet is unknown due to lack of config data for the destination device).

2. The queries are as fast as or faster than manual lookups and joins.

**Screenshots**
N/A

**Additional context**
The speed issue appears to be from usage of `JOIN` commands instead of `LEFT JOIN`.  During testing, changing to `LEFT JOIN` (e.g., for `device_acl_rules`) and then filtering out `NULL` values where reasonable resulted in queries taking significantly less time (e.g., order of 10's of seconds to minutes vs 10's of minutes to hours).

marshall-sg:
This resolves #131 and #130.

---

Specifically, this includes:
- Updated `Route` object to remove logic which sets default values when destination IP or next hop are set.
  - Re-configured validity checks to support known use-cases.
- DB ACL logic updated to perform `LEFT JOIN` instead of `JOIN` to improve operational performance.
- Parser logic updates for `nmdb-import-cisco-show-ip-route`.
  - Supports IOS/ASA/NXOS type formats (very similar, but not exactly the same).
  - Fixed issue with default, empty, object propagation.
- Import tools for Cisco configs regarding routing now flag `Null0` interfaces as a null route (i.e., blackhole route).
- The `nmdb-initialize` tool now performs a DB `TRUNCATE` instead of `DELETE` for improved performance when large amounts of data exists in the DB. 
- Added logic to automate graph generation for DB schema, object, and tool documentation.
  - Primarily improved ability to re-generate when changes occur (i.e., not manual anymore), however more detail has also been added.
  - Generated DB schema graphs for tables only and for tables and views combined.
- Several updates to bring older code into consistent alignment with itself.
  - Clearer function naming, debug output, simpler equality checking, more unit test, etc.


marshall-sg (commented):
Initial content review for change summary.

marshall-sg (commented):
changes since last review seem fine

PR #137, opened 2023-06-09
Datalake branch usage

cmwill:
Datalake now uses the currently selected branch.

### DIFF
--- a/datalake/common/handlers/Git.cpp
+++ b/datalake/common/handlers/Git.cpp
@@ -66,8 +66,11 @@ namespace netmeld::datalake::handlers {
     LOG_DEBUG << "Target time: " << _dts << '\n';
 
     std::ostringstream oss;
+    oss << "git branch --show-current";
+    auto branch {nmcu::trim(nmcu::cmdExecOut(oss.str()))};
+    oss.str("");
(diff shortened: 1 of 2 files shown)


datalake/common/handlers/Git.cpp
```diff
@@ -66,8 +66,11 @@ namespace netmeld::datalake::handlers {
     LOG_DEBUG << "Target time: " << _dts << '\n';
 
     std::ostringstream oss;
+    oss << "git branch --show-current";
```
marshall-sg, line 69: This will only work if the repo is in a defined branch (i.e., not in a detached state).  As such, this conflicts with the remainder of the logic in the code.  Basically, if you run `nmdl-initialize`, check something in (e.g., `nmdl-insert --device-id test /etc/hosts`), and run `nmdl-list` you can see the tool fails.

It might be best/easiest to:
- run the command
- if it returns output stash and use it
- else use the stashed value

marshall-sg (changes requested):
While it does compile, the code doesn't work as currently implemented.  See the code comment.

marshall-sg (approved):
Looks good.

PR #147, opened 2023-10-09
Updates to allow usage of new libpqxx version

marshall-sg:
The change from `libpqxx` version 6.4 to 7.8 required more significant modifications than expected.  Most notably, it currently requires building libpqxx from source and leveraging that during the Netmeld build process.  This seems particularly the case as the one available via package management was built leveraging the C++17 standard while Netmeld code leverages the C++20 standard; and this ultimately causes linking issues during Netmeld's build process.  Rebuilding the libpqxx library with C++20 resolves the issues.

Generally, updates include:
- Modifications to be able to use libpqxx 7.8 (breaking differences from 6.4)
- Realignment of PostgreSQL additional type support logic into Datastore
- Extra error condition checks for `nmdb-initialize` which covers previously unhandled states
- The `nmdb-remove-tool-run` now can handle multiple ids for removal


marshall-sg (commented):
Initial pass review.

PR #148, opened 2023-11-28
111 importer for window systeminfo command

### LINKED ISSUE #111: Importer for Window `systeminfo` command
The windows command `systeminfo` contains quite a lot of information about a host, however there is no tool to import the data.

Create a tool `nmdb-import-systeminfo`(?) that imports the Windows `systeminfo` command output.

Manually inserting the data.

**Additional context**
Sample output:
```
Host Name:                 SomeHostId
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.12345 N/A Build 12345
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Member Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          Some Name                                                                                                                                                          
Registered Organization:   Some Name                                            
Product ID:                00123-00000-00001-AB123                              
Original Install Date:     1/01/1234, 3:59:23 PM                                
System Boot Time:          12/31/1234, 9:43:32 AM                               
System Manufacturer:       CompyOne Inc.                                            
System Model:              Sleek 1234                                        
System Type:               x64-based PC                                         
Processor(s):              1 Processor(s) Installed.                            
                           [01]: Intel64 Family 6 Model 60 Stepping 3 GenuineIntel ~3000 Mhz
BIOS Version:              BInOS Inc. R12, 1/2/2000                              
Windows Directory:         C:\WINDOWS                                           
System Directory:          C:\WINDOWS\system32                                  
Boot Device:               \Device\HarddiskVolume2                              
System Locale:             en-us;English (United States)                        
Input Locale:              en-us;English (United States)                        
Time Zone:                 (UTC-07:00) Mountain Time (US & Canada)              
Total Physical Memory:     16,000 MB                                            
Available Physical Memory: 4,000 MB                                             
Virtual Memory: Max Size:  32,000 MB                                            
Virtual Memory: Available: 18,000 MB                                            
Virtual Memory: In Use:    14,000 MB                                            
Page File Location(s):     C:\pagefile.sys                                      
Domain:                    sub.domain.root                                      
Logon Server:              \\DC012SUBDOM                                        
Hotfix(s):                 8 Hotfix(s) Installed.                               
                           [01]: KB1234567                                      
                           [02]: KB2183345                                      
                           [03]: KB1496911                                      
                           [04]: KB3101010                                      
                           [05]: KB4999123                                      
                           [06]: KB5024632                                      
                           [07]: KB7024532                                      
                           [08]: KB9113471                                      
Network Card(s):           2 NIC(s) Installed.                                  
                           [01]: Intel(R) Ethernet Connection IE12-AB           
                                 Connection Name: Ethernet                      
                                 DHCP Enabled:    Yes                           
                                 DHCP Server:     192.168.111.1                 
                                 IP address(es)                                 
                                 [01]: 192.168.111.101                          
                                 [02]: fe80::dead:beef:feed:fade                
                           [02]: VirtualBox Host-Only Ethernet Adapter          
                                 Connection Name: VirtualBox Host-Only Network  
                                 DHCP Enabled:    No                            
                                 IP address(es)
                                 [01]: 192.168.10.1
                                 [02]: fe80::face:fade:beef:deed
Hyper-V Requirements:      VM Monitor Mode Extensions: Yes
                           Virtualization Enabled In Firmware: Yes
                           Second Level Address Translation: Yes
                           Data Execution Prevention Available: Yes
```

ljvalen:


datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+  system_type =
+      ("System Type: " > token[pnx::bind(&nmdo::DeviceInformation::setDeviceType, &data.devInfo, qi::_1)] > qi::eol);
+
+  domain =
+      ("Domain: " > token > qi::eol); // see notes
+
+  hotfix =
+      +("[" >> qi::ascii::digit >> qi::ascii::digit >> "]:" > token[pnx::bind(&nmdo::OperatingSystem::addHotfix, &data.os, qi::_1)] - qi::omit[qi::eol])
+      ;
+  hotfixs =
```
marshall-sg, line 69: Typo.
```suggestion
  hotfixes =
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+      > os_manufacturer
+      > os_configuration
+      > *(qi::char_ - qi::lit("System Manufacturer:"))
+      > system_manufacturer
+      > system_model
+      > system_type
+      > *(qi::char_ - qi::lit("Domain:"))
+      > domain[pnx::bind(&Parser::setDomain, this, qi::_1)]
+      > *(qi::char_ - qi::lit("Hotfix"))
+      > hotfixs
```
marshall-sg, line 129:
```suggestion
      > hotfixes
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+      (host_name)
+      (os_name)
+      (os_version)
+      (os_manufacturer)
+      (os_configuration)
+      (system_manufacturer)
+      (system_model)
+      (system_type)
+      (domain)
+      (hotfixs)
```
marshall-sg, line 149:
```suggestion
      (hotfixes)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+      (networkCardConnectionName)
+      (networkCardStatus)
+  );
+}
+
+// =============================================================================
+// Parser helper methods
+// =============================================================================
+
+void Parser::setDomain(const std::string &_string)
```
marshall-sg, line 166: Ampersand goes with the type, not the variable name.
```suggestion
void Parser::setDomain(const std::string& _string)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+  curDomain = _string;
+}
+
+void Parser::addHotfix(const std::string &_string)
```
marshall-sg, line 171:
```suggestion
void Parser::addHotfix(const std::string& _string)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+void Parser::addIfaceIp(nmdo::IpAddress &_ipAddr)
+{
+  auto &iface{data.network_cards[curIfaceName]};
+  std::string fqdn = data.devInfo.getDeviceId() + '.' + curDomain;
+  _ipAddr.addAlias(fqdn, "from systeminfo");
+  data.os.setIpAddr(_ipAddr);
+  iface.addIpAddress(_ipAddr);
+}
+
+void Parser::setIfaceStatus(const std::string &_status)
```
marshall-sg, line 204:
```suggestion
void Parser::setIfaceStatus(const std::string& _status)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+  auto &iface{data.network_cards[curIfaceName]};
+  if (_connectionname.find("Wi-Fi") != std::string::npos || _connectionname.find("Bluetooth") != std::string::npos)
+  {
+    iface.setMediaType(_connectionname);
+  }
+  iface.setUp();
+  iface.setDescription(_connectionname);
+}
+
+void Parser::addIfaceIp(nmdo::IpAddress &_ipAddr)
```
marshall-sg, line 195:
```suggestion
void Parser::addIfaceIp(nmdo::IpAddress& _ipAddr)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void Parser::addInterface(const std::string &_name)
+{
+  nmdo::Interface iface;
+  iface.setName(_name);
+  curIfaceName = iface.getName();
+  data.network_cards[curIfaceName] = iface;
+}
+
+void Parser::addIfaceConnectName(const std::string &_connectionname)
```
marshall-sg, line 184:
```suggestion
void Parser::addIfaceConnectName(const std::string& _connectionname)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+  curDomain = _string;
+}
+
+void Parser::addHotfix(const std::string &_string)
+{
+  data.os.addHotfix(_string);
+}
+
+void Parser::addInterface(const std::string &_name)
```
marshall-sg, line 176:
```suggestion
void Parser::addInterface(const std::string& _name)
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+  nmdo::Interface iface;
+  iface.setName(_name);
+  curIfaceName = iface.getName();
+  data.network_cards[curIfaceName] = iface;
+}
+
+void Parser::addIfaceConnectName(const std::string &_connectionname)
+{
+  auto &iface{data.network_cards[curIfaceName]};
+  if (_connectionname.find("Wi-Fi") != std::string::npos || _connectionname.find("Bluetooth") != std::string::npos)
```
marshall-sg, line 187: Too long of a line and can easily miss the or.
```suggestion
  if (  _connectionname.find("Wi-Fi") != std::string::npos
     || _connectionname.find("Bluetooth") != std::string::npos
     )
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+  nmdo::Interface iface;
+  iface.setName(_name);
+  curIfaceName = iface.getName();
+  data.network_cards[curIfaceName] = iface;
+}
+
+void Parser::addIfaceConnectName(const std::string &_connectionname)
+{
+  auto &iface{data.network_cards[curIfaceName]};
```
marshall-sg, line 186:
```suggestion
  auto& iface{data.network_cards[curIfaceName]};
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+  {
+    iface.setMediaType(_connectionname);
+  }
+  iface.setUp();
+  iface.setDescription(_connectionname);
+}
+
+void Parser::addIfaceIp(nmdo::IpAddress &_ipAddr)
+{
+  auto &iface{data.network_cards[curIfaceName]};
```
marshall-sg, line 197:
```suggestion
  auto& iface{data.network_cards[curIfaceName]};
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+  std::string fqdn = data.devInfo.getDeviceId() + '.' + curDomain;
+  _ipAddr.addAlias(fqdn, "from systeminfo");
+  data.os.setIpAddr(_ipAddr);
+  iface.addIpAddress(_ipAddr);
+}
+
+void Parser::setIfaceStatus(const std::string &_status)
+{
+  auto &iface{data.network_cards[curIfaceName]};
```
marshall-sg, line 206:
```suggestion
  auto& iface{data.network_cards[curIfaceName]};
```

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,139 @@
...
+// Data Struct definition
+// =============================================================================
+struct Data
+{
+  auto operator<=>(const Data&) const = default;
+  bool operator==(const Data&) const = default;
+
+  nmdo::DeviceInformation                 devInfo;
+  nmdo::OperatingSystem                   os;
+  std::vector<std::string>                hotfixs;
```
marshall-sg, line 51:
```suggestion
  std::vector<std::string>                hotfixes;
```

Why is this here?  Isn't this handled in `OperatingSystem`?

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,139 @@
...
+      start;
+
+    qi::rule<nmdp::IstreamIter, qi::ascii::blank_type>
+      ignoredLine;
+
+    qi::rule<nmdp::IstreamIter, Systeminfo(), qi::ascii::blank_type>
+      systeminfo;
+
+    qi::rule<nmdp::IstreamIter, std::string(), qi::ascii::blank_type>
+      hotfixs,
```
marshall-sg, line 87:
```suggestion
      hotfixes,
```

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,139 @@
...
+  private:
+    Result getData();
+    void setDomain(const std::string&);
+    void addHotfix(const std::string&);
+    void addInterface(const std::string&);
+    void addIfaceConnectName(const std::string&);
+    void addIfaceIp(nmdo::IpAddress&);
+    void setIfaceStatus(const std::string&);
+};
+#endif // PARSER_HPP
```
marshall-sg, line 139: Please ensure EOL after all lines.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+namespace nmdp = netmeld::datastore::parsers;
+
+using qi::ascii::blank;
+
+class TestParser : public Parser
+{
+  public:
+    using Parser::start;
+    using Parser::hotfix;
+    using Parser::hotfixs;
```
marshall-sg, line 42:
```suggestion
    using Parser::hotfixes;
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+                "Parse rule 'ipAddressLine': " << test);
+    }
+    for (const auto& test: testsFail) {
+      BOOST_TEST(!nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'ipAddressLine fails': " << test);
+    }
+  }
+}
+
+BOOST_AUTO_TEST_CASE(testHotfixs)
```
marshall-sg, line 130:
```suggestion
BOOST_AUTO_TEST_CASE(testHotfixes)
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+{
+  TestParser tp;
+  {
+    const auto& parserRule {tp.hotfixs};
```
marshall-sg, line 134:
```suggestion
    const auto& parserRule {tp.hotfixes};
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+      [10]: KBXXXXXX
+      )STR"
+    };
+    std::vector<std::string> testsFail {
+      R"STR(KBXXXXXX)STR", //Missing index
+      R"STR([01]: )STR"    //Missing Hotfix
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfixs': " << test);
```
marshall-sg, line 155:
```suggestion
                "Parse rule 'hotfixes': " << test);
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+    for (const auto& test: testsFail) {
+      BOOST_TEST(!nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfixs': " << test);
```
marshall-sg, line 159:
```suggestion
                "Parse rule 'hotfixes': " << test);
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+      R"STR([01]: KBXXXXXX)STR"
+    };
+
+    std::vector<std::string> testsFail {
+      R"STR(KBXXXXXX)STR", //Missing index
+      R"STR([01]: )STR"    //Missing Hotfix
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfix': " << test);
```
marshall-sg, line 174:
```suggestion
                "Parse rule 'hotfixes': " << test);
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+    for (const auto& test: testsFail) {
+      BOOST_TEST(!nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfix': " << test);
```
marshall-sg, line 178:
```suggestion
                "Parse rule 'hotfixes': " << test);
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    }
+  }
+}
+
+BOOST_AUTO_TEST_CASE(testWhole)
+{
+  TestParser tp;
+  const auto &parserRule {tp};
```
marshall-sg, line 186:
```suggestion
  const auto& parserRule {tp};
```

marshall-sg (changes requested):
There are numerous changes which are needed, both small and large.  Regardless, with the last edit, the code no longer compiles.  I will re-review once the code compiles again.

datastore/common/objects/OperatingSystem.cpp
```diff
@@ -84,6 +96,26 @@ namespace netmeld::datastore::objects {
         ;
   }
 
+  std::string
+  OperatingSystem::ConvertVectorToPostgresArray(const std::vector<std::string>& vec) {
+    std::string result = "{";
+
+    for (const auto& element : vec) {
+        // Escape special characters and handle NULL elements if needed
+        // You might need additional logic here depending on your data
```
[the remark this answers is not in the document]
ljvalen: Removed

datastore/common/objects/OperatingSystem.hpp
```diff
@@ -63,14 +64,17 @@ namespace netmeld::datastore::objects {
...
       void setProductName(const std::string&);
       void setProductVersion(const std::string&);
       void setCpe(const std::string&);
       void setAccuracy(const double);
+      void setHotfixs(const std::vector<std::string>&);
+      void addHotfix(const std::string&);
 
       bool isValid() const override;
-
+      std::string ConvertVectorToPostgresArray(const std::vector<std::string>& vec);
```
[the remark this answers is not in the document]
ljvalen: The idea was to have the hotfixes not be separate but to be stored in a postgres array. Instead of just a concatenated string. I can move it to utils if we will use it again?

datastore/common/utils/QueriesCommon.cpp
```diff
@@ -950,10 +950,10 @@ namespace netmeld::datastore::utils {
       ("insert_raw_operating_system",
        "INSERT INTO raw_operating_systems AS orig"
        "  (tool_run_id, ip_addr,"
-       "   vendor_name, product_name, product_version, cpe, accuracy)"
+       "   vendor_name, product_name, product_version, cpe, accuracy, hotfixs)"
        " VALUES ($1, host(($2)::INET)::INET,"
        "         nullif($3, ''), nullif($4, ''), nullif($5, ''),"
-       "         nullif($6, ''), $7)"
+       "         nullif($6, ''), $7, $8)"
```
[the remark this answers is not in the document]
ljvalen: Moved into  its own query for hotfixes.

datastore/importers/nmdb-import-systeminfo/README.md
```diff
@@ -0,0 +1,33 @@
...
+remote host and displays the results locally, then the following would
+process it and save the data to a file called `systeminfoOutput.txt` in the current
+working directory.
+```
+... | nmdb-import-systeminfo --pipe --device-id <workstation> packages.txt
+```
+TROUBLESHOOTING
+========
+
+There is trouble with the actual output from systeminfo. It is unicode 16 little endian with crlf so you can convert it with `iconv -f UTF-16 -t US-ASCII ~/systeminfo.txt -o ~/systeminfoascii.txt` and then I had to convert it using `uni2ascii` and then I had to remove all the '\r' with `cat systeminfoascii2.txt | tr -d '\r' > ~/systeminfoascii3.txt` and then I had to remove the first line which was a /n `cat ~/systeminfoascii.txt | tail -n +2 > systeminfoascii2.txt`
```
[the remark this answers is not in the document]
ljvalen: Moved to Description and rewritten.

datastore/importers/nmdb-import-systeminfo/README.md
```diff
@@ -0,0 +1,33 @@
+DESCRIPTION
+===========
+
+Parse and import the output of the
+`systeminfo`
+command on Windows
+systems into the Netmeld framework for later analysis. Import's output
+based on device-id.
+
```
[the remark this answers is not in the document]
ljvalen: I rewrote to include the steps

datastore/common/objects/OperatingSystem.cpp
```diff
@@ -84,6 +96,26 @@ namespace netmeld::datastore::objects {
         ;
   }
 
+  std::string
+  OperatingSystem::ConvertVectorToPostgresArray(const std::vector<std::string>& vec) {
```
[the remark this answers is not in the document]
ljvalen: Do we want to store the hotfixes in a concatenated string? If so that will work but I thought initially we could store them in an array, to have more flexibility over retrieval?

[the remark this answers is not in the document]
ljvalen: https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT

datastore/common/objects/OperatingSystem.hpp
```diff
@@ -63,14 +64,17 @@ namespace netmeld::datastore::objects {
     private:
     protected:
     public:
+      void setIpAddr(const IpAddress&);
       void setVendorName(const std::string&);
       void setProductName(const std::string&);
       void setProductVersion(const std::string&);
       void setCpe(const std::string&);
       void setAccuracy(const double);
+      void setHotfixs(const std::vector<std::string>&);
```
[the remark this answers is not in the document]
ljvalen: Removed

datastore/common/schemas/022tool-results-views-create.sql
```diff
@@ -649,7 +649,8 @@ SELECT DISTINCT
     product_name                AS product_name,
     product_version             AS product_version,
     cpe                         AS cpe,
-    accuracy                    AS accuracy
+    accuracy                    AS accuracy,
+    hotfixs                     AS hotfixs
```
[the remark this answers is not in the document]
ljvalen: Outdated

datastore/common/utils/QueriesCommon.cpp
```diff
@@ -950,10 +950,10 @@ namespace netmeld::datastore::utils {
       ("insert_raw_operating_system",
        "INSERT INTO raw_operating_systems AS orig"
        "  (tool_run_id, ip_addr,"
-       "   vendor_name, product_name, product_version, cpe, accuracy)"
+       "   vendor_name, product_name, product_version, cpe, accuracy, hotfixs)"
```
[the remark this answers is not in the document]
ljvalen: Changed this functionality

datastore/importers/nmdb-import-systeminfo/Parser.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.hpp
ljvalen: Seperating hotfixes into seperate table show in the operating system view

datastore/importers/nmdb-import-systeminfo/Parser.hpp
ljvalen: Fixed.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Fixed

datastore/common/schemas/032aws-views-create.sql
```diff
@@ -613,4 +613,4 @@ LEFT JOIN aws_network_acl_rules_joined AS t2
 -------------------------------------------------------------------------------
 
 
-COMMIT TRANSACTION;
+COMMIT TRANSACTION;
```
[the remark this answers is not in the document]
marshall-sg: I reverted this change.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+      )STR"
+    };
+    std::vector<std::string> testsFail {
+      R"STR(KBXXXXXX)STR", //Missing index
+      R"STR([01]: )STR"    //Missing Hotfix
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfixes': " << test);
+    }
```
marshall-sg, line 156: While you should add a check since state was changed, you don't need to check that each hotfix was added since that logic is technically in another rule.  Just check that the hotfix count is not zero (you probably also need to clear the `hotfixes` vector between runs).

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    };
+
+    std::vector<std::string> testsFail {
+      R"STR(KBXXXXXX)STR", //Missing index
+      R"STR([01]: )STR"    //Missing Hotfix
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'hotfix': " << test);
+    }
```
marshall-sg, line 175: State should be changed.  You need a test to ensure the hotfix was updated with what was parsed.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+                                  [02]: 2001:0db8:85a3:0000:0000:8a2e:0370:7334
+          Hyper-V Requirements:      A hypervisor has been detected. Features required for Hyper-V will not be displayed.
+
+          )STR"
+   };
+
+  for (const auto& test : testsOk) {
+    BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+              "Parse rule 'start': " << test);
+  }
```
marshall-sg, line 252: If you check state change validity at each rule, you shouldn't really need it here.  You can put it if you want, but it can get complicated if you have multiple formats you want to check.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        iface.setMediaType(_connectionname);
+    }
+    iface.setUp();
+    iface.setDescription(_connectionname);
+}
+
+void Parser::addIfaceIp(nmdo::IpAddress &_ipAddr)
+{
+    auto &iface{data.network_cards[curIfaceName]};
+    std::string fqdn = data.devInfo.getDeviceId() + '.' + curDomain;
```
marshall-sg, line 149: Prefer initialization over assignment (saves a copy operation at least).  Also prefer list initialization, there are additional guarantees (e.g., doesn't allow narrowing).  It's not always the better or possible choice, but it typically will be.

```suggestion
    std::string fqdn {data.devInfo.getDeviceId() + '.' + curDomain};
```

datastore/common/objects/OperatingSystem.hpp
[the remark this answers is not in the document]
marshall-sg: No, don't move it to utils.  It ultimately is unnecessary logic.

Regardless, I am marking this as resolved since it is outdated and will comment on the new code.

datastore/common/objects/OperatingSystem.cpp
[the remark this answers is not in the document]
marshall-sg: No we don't want a concatenated string.  Also, not sure why you are referencing PostgreSQL documentation in this case as that's not the layer you are working at.  It is libpqxx documentation you should be referencing (as I mentioned in my prior comment).

Regardless, I am marking this as resolved since it is outdated and will comment on the new code.

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+          result += "'" + element + "',";
+      }
+
+      if (!vec.empty()) {
+          result.pop_back();
+      }
+
+      result += "}";
+      return result;
+    }
```
marshall-sg, line 110: Again, do not do this.
1. The logic is wrong.  You are adding single quotes and, thus, changing the hotfix value.  So any search is now `'''KBXXXXXX'''` instead of just `'KBXXXXXX'`.  As I mentioned before you are working with libpqxx not a Postgresql command directly.
2. I just tested it, nothing is needed besides just inserting the vector.  The libpqxx calls handle everything for you.

So this whole function can go away.

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+        LOG_DEBUG << "Iterating over DeviceInformation\n";
+        results.devInfo.save(t, toolRunId, deviceId);
+        LOG_DEBUG << results.devInfo.toDebugString() << "\n";
+
+        LOG_DEBUG << "Iterating over OperatingSystems\n";
+        results.os.save(t, toolRunId, deviceId);
+        LOG_DEBUG << results.os.toDebugString() << "\n";
+
+        LOG_DEBUG << "Iterating over Hotfixes\n";
+        t.exec_prepared("insert_raw_hotfixes", toolRunId, ConvertVectorToPostgresArray(results.hotfixes));
```
marshall-sg, line 79: See my below comment, but this is much more complicated than needed.
```suggestion
        t.exec_prepared("insert_raw_hotfixes", toolRunId, results.hotfixes);
```

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+    _ipAddr.addAlias(fqdn, "from systeminfo");
+    data.os.setIpAddr(_ipAddr);
```
marshall-sg, line 151: Couple of thoughts:
1. This overrides the IP address to whatever is last processed.  This is less than ideal as the last processed is probably an IPv6 address that isn't really useful (e.g., `fe80:...`).
2. It may not be the primary interface.
3. No interface may have an IP (e.g., if systeminfo is ran while the network is down).

All of this is less than ideal.  Please set up some time to discuss this as it may take reworking the `OperatingSystem` object.

datastore/common/objects/OperatingSystem.cpp
```diff
@@ -94,8 +100,9 @@ namespace netmeld::datastore::objects {
                 << std::endl;
       return;
     }
-
-    ipAddr.save(t, toolRunId, deviceId);
+    if (ipAddr.isValid()) { 
+      ipAddr.save(t, toolRunId, deviceId);
+    }
```
marshall-sg, line 105: The IP address has to be valid when reaching here, otherwise the logic short circuits prior to ever reaching this point.  Also, the `save()` method already does a validity check prior to doing anything.  So this extra check is, thrice, redundant and not needed here.

marshall-sg (changes requested):
Rework is required.  See my comments.

Not sure why it split my comments.  Both sets are/were part of the same review.

marshall-sg:
Also, I merged the `origin/master` branch as your code was/is considerably out of sync.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        ("System Model: " > token[pnx::bind(&nmdo::DeviceInformation::setModel, &data.devInfo, qi::_1)] > qi::eol);
+
+    system_type =
+        ("System Type: " > token[pnx::bind(&nmdo::DeviceInformation::setDeviceType, &data.devInfo, qi::_1)] > qi::eol);
+
+    domain =
+        ("Domain: " > token > qi::eol); 
+
+    hotfix =
+        +("[" >> qi::ascii::digit >> qi::ascii::digit >> "]:" > token[pnx::bind(&Parser::addHotfix, this, qi::_1)] - qi::omit[qi::eol]);
```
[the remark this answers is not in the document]
ljvalen: fixed this

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,137 @@
...
+      os_version,
+      os_manufacturer,
+      os_configuration,
+      system_manufacturer,
+      system_model,
+      system_type,
+      domain,
+      hotfix,
+      hyper_v,
+      token;
```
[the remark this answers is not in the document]
ljvalen: Cleaned up the header to only have the necessary rules

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+Parser::Parser() : Parser::base_type(start)
+{
+    start =
+        (*qi::eol > systeminfo > -qi::eol)[(qi::_val = pnx::bind(&Parser::getData, this))];
+
+    token =
+        +qi::ascii::graph;
+
+    host_name =
+        ("Host Name: " > token[pnx::bind(&nmdo::DeviceInformation::setDeviceId, &data.devInfo, qi::_1)] > qi::eol);
```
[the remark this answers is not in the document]
ljvalen: Ahh okay, I figured I would take the approach of just parsing until we the match and then grab the information then and continue the rest of the rule, If its okay with you, I have the parser working and I would end up spending more time trying to troubleshoot changing all these rules

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        ('[' >> qi::lexeme[+qi::char_("0-9")] >> ']' >> qi::lit(':') >> networkCardName[(pnx::bind(&Parser::addInterface, this, qi::_1))] >> qi::eol >> "Connection Name: " > networkCardConnectionName[(pnx::bind(&Parser::addIfaceConnectName, this, qi::_1))] >> qi::eol >> -(dhcpEnabledStatus >> qi::eol) >> -("DHCP Server: " >> dhcpServer >> qi::eol) >> -("IP address(es)" >> qi::eol > +('[' >> qi::lexeme[+qi::char_("0-9")] >> "]: " >> ipAddr[(pnx::bind(&Parser::addIfaceIp, this, qi::_1))] >> qi::eol)) >> -("Status: " >> networkCardStatus[(pnx::bind(&Parser::setIfaceStatus, this, qi::_1))] >> qi::eol));
+    ;
+    network_cards =
+        qi::lit("Network Card(s):") >> +qi::ascii::print > qi::eol > +network_card;
+
+    hyper_v =
+        ("Hyper-V Requirements: " > token > qi::eol);
+
+    systeminfo =
+        host_name > os_name > os_version > os_manufacturer > os_configuration > *(qi::char_ - qi::lit("System Manufacturer:")) > system_manufacturer > system_model > system_type > *(qi::char_ - qi::lit("Domain:")) > domain[pnx::bind(&Parser::setDomain, this, qi::_1)] > *(qi::char_ - qi::lit("Hotfix")) > hotfixes > network_cards > hyper_v;
```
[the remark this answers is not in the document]
ljvalen: Okay I fixed the formatting and removed the setdomain function

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        +qi::ascii::graph;
+
+    network_card =
+        ('[' >> qi::lexeme[+qi::char_("0-9")] >> ']' >> qi::lit(':') >> networkCardName[(pnx::bind(&Parser::addInterface, this, qi::_1))] >> qi::eol >> "Connection Name: " > networkCardConnectionName[(pnx::bind(&Parser::addIfaceConnectName, this, qi::_1))] >> qi::eol >> -(dhcpEnabledStatus >> qi::eol) >> -("DHCP Server: " >> dhcpServer >> qi::eol) >> -("IP address(es)" >> qi::eol > +('[' >> qi::lexeme[+qi::char_("0-9")] >> "]: " >> ipAddr[(pnx::bind(&Parser::addIfaceIp, this, qi::_1))] >> qi::eol)) >> -("Status: " >> networkCardStatus[(pnx::bind(&Parser::setIfaceStatus, this, qi::_1))] >> qi::eol));
+    ;
+    network_cards =
+        qi::lit("Network Card(s):") >> +qi::ascii::print > qi::eol > +network_card;
+
+    hyper_v =
+        ("Hyper-V Requirements: " > token > qi::eol);
```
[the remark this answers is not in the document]
ljvalen: We discussed this and decided we don't need the hyper-v information so I omitted everything after network-cards

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        qi::lexeme[+(qi::char_ - qi::eol)];
+    ;
+    dhcpEnabledStatus =
+        qi::lit("DHCP Enabled:") >> (qi::lit("Yes") | qi::lit("No"));
+
+    networkCardStatus =
+        +qi::ascii::graph;
+
+    network_card =
+        ('[' >> qi::lexeme[+qi::char_("0-9")] >> ']' >> qi::lit(':') >> networkCardName[(pnx::bind(&Parser::addInterface, this, qi::_1))] >> qi::eol >> "Connection Name: " > networkCardConnectionName[(pnx::bind(&Parser::addIfaceConnectName, this, qi::_1))] >> qi::eol >> -(dhcpEnabledStatus >> qi::eol) >> -("DHCP Server: " >> dhcpServer >> qi::eol) >> -("IP address(es)" >> qi::eol > +('[' >> qi::lexeme[+qi::char_("0-9")] >> "]: " >> ipAddr[(pnx::bind(&Parser::addIfaceIp, this, qi::_1))] >> qi::eol)) >> -("Status: " >> networkCardStatus[(pnx::bind(&Parser::setIfaceStatus, this, qi::_1))] >> qi::eol));
```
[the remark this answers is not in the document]
ljvalen: Okay I line broke at expects, the formatter I used was making them 1 liners

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,167 @@
...
+        ("System Type: " > token[pnx::bind(&nmdo::DeviceInformation::setDeviceType, &data.devInfo, qi::_1)] > qi::eol);
+
+    domain =
+        ("Domain: " > token > qi::eol); 
+
+    hotfix =
+        +("[" >> qi::ascii::digit >> qi::ascii::digit >> "]:" > token[pnx::bind(&Parser::addHotfix, this, qi::_1)] - qi::omit[qi::eol]);
+    
+    hotfixes =
+        ("Hotfix(s): " >> (+qi::ascii::print - qi::eol) >> qi::eol > *(hotfix > qi::eol));
```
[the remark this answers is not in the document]
ljvalen: Do you have any reference to this? Otherwise I don't think I know how I would validate if a hotfix is valid

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,137 @@
...
+      systeminfo;
+
+    qi::rule<nmdp::IstreamIter, std::string(), qi::ascii::blank_type>
+      hotfixes,
+      networkCardName,
+      networkCardConnectionName,
+      dhcpServer,
+      networkCardStatus,
+      ipAddressLine
+    ;
```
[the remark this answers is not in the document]
ljvalen: I adjusted these in an earlier push

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+    qi::rule<nmdp::IstreamIter, std::vector<std::string>(), qi::ascii::blank_type>
+      ipaddresssection
+    ;
```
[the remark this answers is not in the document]
ljvalen: addressed

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+    qi::rule<nmdp::IstreamIter, bool(), qi::ascii::blank_type>
+      dhcpEnabledStatus;
```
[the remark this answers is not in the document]
ljvalen: I must've associated return values with the semantic actions

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+class TestParser : public Parser
+{
+  public:
+    using Parser::start;
+    using Parser::hotfix;
+    using Parser::hotfixes;
+    using Parser::network_cards;
+    using Parser::network_card;
+    using Parser::ipAddressLine;
```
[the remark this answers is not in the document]
ljvalen: Alot of the rules don't actually have any "ruling" only those ones have logic within the grammer, most others are just tokens

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+BOOST_AUTO_TEST_CASE(testNetworkCard)
+{
+  TestParser tp;
+  {
+    const auto& parserRule {tp.network_card};
+    std::vector<std::string> testsOk {
+      R"STR([01]: Realtek PCIe GbE Family Controller
+      Connection Name: Ethernet
+      DHCP Enabled:    Yes
```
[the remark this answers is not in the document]
ljvalen: Okay added a test

datastore/importers/nmdb-import-systeminfo/Parser.cpp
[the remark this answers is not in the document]
marshall-sg: Sure.  As I said, you don't have to make the change.

Just of note, using semantic actions can cause problems and, your described usage case, ultimately may make you run into more problems than necessary.  This is primarily because of the state change they cause.  Luckily, you aren't allowing backtracking so there isn't much of a concern as if the rule logic fails, the whole parser fails.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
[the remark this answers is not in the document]
marshall-sg: We talked offline on this topic.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+      [01]: 192.168.0.2
+      [02]: 2001:0db8:85a3:0000:0000:8a2e:0370:7334
+      )STR", //Missing ipaddresses label
+      R"STR([01]: Realtek PCIe GbE Family Controller
+      Connection Name: Ethernet
+      DHCP Enabled:    Yes
+      IP address(es)
+      )STR" //Missing Ipaddress data
+    };
```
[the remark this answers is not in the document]
ljvalen: My thought behind these was to ensure that the more complex rules actually get parsed? Is that a valid reason to keep them?

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+      [03]: KBXXXXXX
+      [04]: KBXXXXXX
+      [05]: KBXXXXXX
+      [06]: KBXXXXXX
+      [07]: KBXXXXXX
+      [08]: KBXXXXXX
+      [09]: KBXXXXXX
+      [10]: KBXXXXXX
+      )STR"
+    };
```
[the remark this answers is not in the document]
ljvalen: Okay I added a case to simulate the buffer exhaustion. The true exhuastion is apparently around 200 hotfixes on some OS's

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    std::vector<std::string> testsFail {
+      R"STR(KBXXXXXX)STR", //Missing index
+      R"STR([01]: )STR"    //Missing Hotfix
+    };
```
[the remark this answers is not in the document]
ljvalen: Okay I just removed them

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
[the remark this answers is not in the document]
marshall-sg: I must be misunderstanding something then.  The object name is `testsFail` so I am assuming these are cases when the parser logic (i.e., `network_card`) should fail (and it is used as such in lines 100-103).  Thus, the text should not actually be consumed or other data objects populated.  So, I'm not understanding how it helps ensure the more complex rules actually get parsed.  I'm only seeing how it helps ensure a dev does not wrongly change the rule and succeed on something the parser should not.

Regardless, like I said it can be kept.  I was mainly pointing out there may be better alternatives depending on your objective and enumerating all the fail cases is not always needed.

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
ljvalen: Okay removed

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
ljvalen: Committed suggestion

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: Okay I added the clear hotfix

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,238 @@
...
+        > qi::eol); 
+
+    hotfix =
+        ("[" >> +qi::ascii::digit >> "]:" >> qi::as_string[qi::ascii::string("KB") >> +qi::ascii::digit][pnx::bind(&Parser::addHotfix, this, qi::_1)]);
+
+    hotfixes =
+        ("Hotfix(s): " 
+        >> (+qi::ascii::print - qi::eol) 
+        >> qi::eol
+        > ( *(hotfix > qi::eol) | ignoredLine));
```
marshall-sg, line 91: I don't think this will work as expected, it probably should be one or more (i.e., `+`).  Saying zero or more means the logic is a success even when the rule fails to parse anything; so it doesn't actually ever consume the bad line and the following rules will fail.

marshall-sg (changes requested):
Changes are needed.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,238 @@
...
+        +qi::ascii::graph;
+
+    hostName =
+        ("Host Name: " 
+        > token[pnx::bind(&nmdo::DeviceInformation::setDeviceId, &data.devInfo, qi::_1)] 
+        > qi::eol);
+
+    osName =
+        ("OS Name: " 
+        > token[pnx::bind(&nmdo::OperatingSystem::setProductName, &data.os, qi::_1)][pnx::bind(&nmdo::OperatingSystem::setCpe, &data.os, qi::_1)]
```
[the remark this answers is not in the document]
ljvalen: Okay addressed this, It shouldn't not succeed since its just setting the name?

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,238 @@
...
+// =============================================================================
+// Parser logic
+// =============================================================================
+Parser::Parser() : Parser::base_type(start)
+{
+    start =
+        (*qi::eol > systemInfo > -qi::eol)[(qi::_val = pnx::bind(&Parser::getData, this))];
+
+    token =
+        +qi::ascii::graph;
```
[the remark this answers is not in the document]
ljvalen: Ahh yeah, I think I did that because of the CPE stuff. Speaking of which, Technically we aren't storing the cpe properly do you want me to create a function to compile the OS's information into a cpe?

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+class TestParser : public Parser
+{
+  public:
+    using Parser::start;
+    using Parser::hotfix;
+    using Parser::hotfixes;
+    using Parser::networkCards;
+    using Parser::networkCard;
+    using Parser::data;
```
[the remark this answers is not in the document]
ljvalen: Yeah this makes sense, I feel that this is more of a syntax problem? I should have known that the token wasn't consuming blanks, Should I write a test for every single use case of token?

datastore/importers/nmdb-import-systeminfo/Parser.hpp
```diff
@@ -0,0 +1,128 @@
...
+  private:
+    Result getData();
+    void addHotfix(const std::string&);
+    void addInterface(const std::string&);
+    void addIfaceConnectName(const std::string&);
+    void addIfaceIp(nmdo::IpAddress&);
+    void setIfaceStatus(const std::string&);
+  public:
+    void clearHotfixes();
+    void clearNetworkCards();
```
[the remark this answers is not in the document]
ljvalen: Yeah actually, I left them by accident, removed them!

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+      R"STR([01]: Realtek PCIe GbE Family Controller
+      Connection Name: Ethernet
+      DHCP Enabled:    Yes
+      IP address(es)
+      )STR" //Missing Ipaddress data
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'networkCard': " << test);
+    }
```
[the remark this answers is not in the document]
ljvalen: Okay yes, I see, I just tested with a debug string and validated the ip address let me know if that works?

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+      [07]: KB000006
+      [08]: KB000007
+      [09]: KB000008
+      [10]: KB000009
+      )STR"
+
+    };
+    for (const auto& test : testsOk) {
+      std::string out;
+      // This test case is special because it is testing the buffer exhaustion
```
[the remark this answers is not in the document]
ljvalen: Yes great catch! Ill remov, and others like it!

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'networkCard': " << test);
+    }
+    for (const auto& test: testsFail) {
+      BOOST_TEST(!nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'networkCard fails': " << test);
+    }
+    tp.clearNetworkCards();
```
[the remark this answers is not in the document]
ljvalen: The information in the tp.network_cards was being persisted through the Test cases so hence the clear, to ensure that each run was unique.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+      Connection Name: Ethernet
+      DHCP Enabled:    Yes
+      IP address(es)
+      [01]: 192.168.0.2
+      [02]: 2001:0db8:85a3:0000:0000:8a2e:0370:7334
+      )STR",
+    };
+    BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank),
+              "Parse rule 'networkCard': " << test);
+    std::string curInterfaceName = "Realtek PCIe GbE Family Controller";
```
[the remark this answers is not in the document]
ljvalen: Indeed. Removed!

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,275 @@
...
+    std::cout << "Interface Name: " << tp.data.network_cards << std::endl;
```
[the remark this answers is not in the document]
ljvalen: This is more developed debug and not actual necessary debug lol! Removed

datastore/importers/nmdb-import-systeminfo/Parser.cpp
```diff
@@ -0,0 +1,238 @@
...
+    data.network_cards[curIfaceName] = iface;
+}
+
+void Parser::addIfaceConnectName(const std::string &_connectionname)
+{
+    auto &iface{data.network_cards[curIfaceName]};
+    if (_connectionname.find("Wi-Fi") != std::string::npos || _connectionname.find("Bluetooth") != std::string::npos)
+    {
+        iface.setMediaType(_connectionname);
+    }
```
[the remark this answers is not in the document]
ljvalen: Wait this is weird, how is the mediaType set to ethernet then if I only have the set function in this conditional? It actually doesn't.... why?? no clue...
[name: intel(r) wi-fi 6 ax200 160mhz, description: Wi-Fi, isUp: false, mediaType: wi-fi, macAddr: [macAddress: Invalid MAC, ipAddrs: [], isResponding: false]]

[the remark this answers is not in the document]
marshall-sg: For `InterfaceNetwork`, the `mediaType` variable is defaulted to `ethernet`.  The problem I'm talking about is if one of "Wi-Fi" or "Bluetooth" is present with more text it will set `mediaType` to the whole text.  The value of `_connectionname` is the rest of the line after `Connection Name:` and that remainder can be several words long.

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+        LOG_DEBUG << "Iterating over OperatingSystems\n";
+        results.os.save(t, toolRunId, deviceId);
+        LOG_DEBUG << results.os.toDebugString() << "\n";
+
+        LOG_DEBUG << "Iterating over Hotfixes\n";
+        t.exec_prepared("insert_raw_hotfixes", toolRunId, results.hotfixes);
+        for (auto& hotfix : results.hotfixes) {
+          LOG_DEBUG << hotfix << "\n";
+        }
```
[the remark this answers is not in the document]
marshall-sg: We can still keep the hotfix data separate in the DB, how you say.  It's just the logic to do the save action (i.e., the `t.exec_prepared()`) and the hotfix vector moves to the `OperatingSystem` object.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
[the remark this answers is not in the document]
ljvalen: Okay I added one!

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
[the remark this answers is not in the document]
ljvalen: The data should remain consistent, so there's no need to write a rule for every test, as long as the rule doesn't perform any hidden actions before being parsed.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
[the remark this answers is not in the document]
ljvalen: Right, But look at the example I pasted in the previous comment? The actual interface was intel(r) wi-fi 6 ax200 160mhz, but the media-type got set to wi-fi

datastore/importers/nmdb-import-systeminfo/nmdb-import-systeminfo.cpp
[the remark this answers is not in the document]
ljvalen: Ohh okay, check now I added a conditional to ensure that the OS object is valid before saving the hotfixes

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,276 @@
...
+      [06]: KB000005
+      [07]: KB000006
+      [08]: KB000007
+      [09]: KB000008
+      [10]: KB000009
+      )STR"
+
+    };
+
+    BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank, false),
```
marshall-sg, line 162: This test should fully consume the input.
```suggestion
    BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
```

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,276 @@
...
+      // Simulate buffer exhaustion
+      R"STR(Hotfix(s):           3 Hotfix(s) Installed.
+      [01]: KB000001
+      [02]: KB000002
+      [03]: KB000003
+      [04]: K
+      )STR"
+    };
+    std::vector<std::string> out;
+    BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank, false),
```
marshall-sg, line 183: Now that I've looked at the whole logic, this test should fully consume the input because you actually want to get rid of the partial line.  Otherwise the rest of the rules won't match correctly (e.g., `networkCards`) when they should because there is still bad data in the buffer.
```suggestion
    BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank),
```

marshall-sg (changes requested):
Closer, but still some issues.

Also, don't forget to resolve the previously called out issues as well.

datastore/importers/nmdb-import-systeminfo/Parser.cpp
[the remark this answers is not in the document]
ljvalen: Okay I just added some if statements to properly set the media type

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
ljvalen: The only problem is I get an error because it couldn't parse the text properly? The false is what fixed the error because it allowed the parser to fail the parse on that line which is expected?

marshall-sg: We talked about this out of band, just putting summary notes for reference.

The failure is expected in the `hotfix` rule, but the `hotfixes` rule needs to completely consume the section labeled "Hotfix(s)".  So, your rules are wrong.

That is, the `systemInfo` rule contains `... > -hotfixes > -networkCards > qi::omit[...`, so if the read buffer still has the next data as `[04]: K`, then `networkCards` won't match and ultimately it will drop all data after the `[04]: K` line since the `qi::omit[]` rule is designed to consume till the end of the input.

ljvalen: Okay addressed this and the tester is working properly. Removed ignoredLine.

datastore/importers/nmdb-import-systeminfo/Parser.test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    };
+    for (const auto& test : testsOk) {
+      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'ipAddressLine': " << test);
+    }
+    for (const auto& test: testsFail) {
+      BOOST_TEST(!nmdp::test(test.c_str(), parserRule, blank),
+                "Parse rule 'ipAddressLine fails': " << test);
+    }
+  }
```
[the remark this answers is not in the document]
ljvalen: removed.

core/common/utils/StringUtilities.cpp
```diff
@@ -135,4 +135,26 @@ namespace netmeld::core::utils {
...
+  std::string
+  compileCPE(const std::string& _part, const std::string& _vendor,
+             const std::string& _product, const std::string& _version)
+  {
+    // CPE format follows NMAP standard: 
+    // cpe:/{part}:{vendor}:{product}:{version}
+
+    std::string cpe {"cpe:/"};
+    cpe += _part + ":" + _vendor + ":" + _product + ":" + _version;
+    return cpe;
```
marshall-sg, line 148: FWIW (and not a required change), this seems like it would be much clearer if you used `std::format()` instead.  So something like.
```suggestion
    return std::format("cpe:/{}:{}:{}:{}", _part, _vendor, _product, _version);
```

marshall-sg (approved):
Looks good.

PR #149, opened 2023-12-13
added check for root user

raulcruise:
Ansible will now automatically gather privileged commands when the user is root without requiring password when default.remote_user is root and using SSH keys.


marshall-sg (changes requested):
The one changed file looks fine.

Please also add commented out versions of the configuration lines mentioned in the associated open issue to the `ansible.cfg` file as well, along with a short note towards their usage.  I believe they are also needed to make this logic work.

marshall-sg (approved):
Looks good.

PR #150, opened 2023-12-22
Prowler v3 mods

marshall-sg:
This PR includes:
- Updates to handle changes between Prowler v2 and v3 output.
  - The same tooling handles both (defaults to v3) and logic combines both for analysis.
- Cleanup of `nmdb-export-scans` to standardize the logic flow for exporting targeted scan data.
- Cleanup of sporadic trailing white space.


marshall-sg (commented):
Initial pass.

PR #154, opened 2024-01-22
added dbPrepareCommon function within nmdb-initialize so that the hum…

raulcruise:
This change allows us to run the decoder tools instantly after running nmdb-initialize without running into a "violates foreign key constraint" error.


marshall-sg (approved):
Looks good.

PR #155, opened 2024-01-25
128 support for juniper command show arp

### LINKED ISSUE #128: Support for Juniper command show arp
The Datastore module provides support to parse `show arp` output for several formats, but Juniper's doesn't appear to be one.  Add support for it.

Add support for Juniper's `show arp` output format.

Manual insertion of the data.

**Additional context**
Unclear if this needs to be an additional tool, an enhancement to the `nmdb-import-show-neighbor` tool (since it already says it supports the `show arp` command), or something else.

ben-anthony-snl:
Closes #128


ben-anthony-snl:
Should be good to go. @marshall-sg can you review this when you have time?

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace nmdp = netmeld::datastore::parsers;
+
+using qi::ascii::blank;
+
+class TestParser : public Parser {
+    public:
+      using Parser::arpJuniperConfig;
+};
+
+BOOST_AUTO_TEST_CASE(testParts)
```
marshall-sg, line 44: There are a couple of issues here.
1. The `arpJuniperConfig` test logic really is just doing a parser pass/fail and, in that sense, is not really different than the `testWhole` logic.  It's not checking anything about the object(s) that should be created and how.  However, anything else may be more complicated, so see point 2.
2. This should have the other parts as well (and prior to, so as to build up to, the full logic).  Specifically, it should have test logic for `arpHeaderJuniper`, `arpEntryJuniper`, and `arpFooterJuniper`.  If that's there, then it will allow for testing of logic piece-wise and make it easier to check whether a particular logic flow (e.g., whether the "Name" column exists or not) behaves correctly.

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
```diff
@@ -0,0 +1,84 @@
+// =============================================================================
```
marshall-sg, line 1: Thanks for making this.  Since there are other parts you are not focusing on, I will provide additional sample cases and logic for the other parts.

datastore/importers/nmdb-import-show-neighbor/Parser.cpp
```diff
@@ -199,6 +200,45 @@ Parser::Parser() : Parser::base_type(start)
        pnx::bind(&nmdo::InterfaceNetwork::addReachableMac, &qi::_val, qi::_b))]
     ;
 
+  // =============================================================================
+  // ARP: Juniper Config
+  // =============================================================================
+
+  arpJuniperConfig =
+    arpHeaderJuniper >>
```
marshall-sg, line 208: I realize the rest of the parser's logic doesn't do what I'm about to state (since it was made prior to the decision), but parser logic should only use the sequence operator, `>>`, when the ability to backtrack is explicitly wanted.  Prefer to use the expectation operator, `>`, in most cases as it is, overall, more explicit and efficient for our purposes.  As such, the majority of this new code should be switched.

marshall-sg (changes requested):
Overall, the logic and test cases need work before it can be accepted.

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+// Copyright 2022 National Technology & Engineering Solutions of Sandia, LLC
```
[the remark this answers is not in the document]
marshall-sg: I fixed this in my push.

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
marshall-sg: I've pushed the additional test logic for the parts not the focus of this PR.

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
ben-anthony-snl: Let me know if the new tests are good enough to cover these. I tried to order them so it could test for Name or no Name and parse the parts correctly. If there's a better way, let me know.

marshall-sg (changes requested):
A couple of changes to the test logic.  One requires a, hopefully, small change to the parser logic.

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
```diff
@@ -39,30 +39,329 @@ using qi::ascii::blank;
...
     };
     for (const auto& test : testsOk) {
-      BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank),
-                 "Parse rule 'arp': " << test);
+        nmdo::InterfaceNetwork out;
+      BOOST_TEST(nmdp::testAttr(test.c_str(), parserRule, out, blank),
+                 "Parse rule 'arpEntryJuniper': " << test);
+      BOOST_TEST(fin == out);
     }
   }
```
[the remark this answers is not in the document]
ben-anthony-snl: Oh yeah. Good to know. I didn't test it because it seemed so straightforward. I didn't realize there were variations for ti

datastore/importers/nmdb-import-show-neighbor/Parser.test.cpp
```diff
@@ -39,30 +39,329 @@ using qi::ascii::blank;
...
+
+      std::vector<std::string> testsOk {
+          "MAC Address       Address         Name                     Interface"
+      };
+      for (const auto& test : testsOk) {
+          BOOST_TEST(nmdp::test(test.c_str(), parserRule, blank)
+                  , "Parse rule 'arpHeaderJuniper': " << test
+                  );
+      }
+  }
```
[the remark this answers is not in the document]
ben-anthony-snl: Let me know if the most recent changes fix this. I believe I did it properly

ben-anthony-snl:
I updated the juniper arp stuff, made some changes to parser rollbacks, and updated ndp ipv6 stuff. Take a look when you get a chance @marshall-sg

marshall-sg (approved):
Looks good.

Issue #141, opened 2023-07-24
Arista show inventory support


marshall-sg:
**Is your feature request related to a problem? Please describe.**
The current `nmdb-import-show-inventory` is confusing.  Other vendors have a `show inventory` command, but it seems the tooling only supports Cisco's version, currently.

**Describe the solution you'd like**
It is unclear if a new tool is needed or an existing tool can be modified to support it, but support for Arista's `show inventory` output is requested.  If a new tool is created instead of modification, please updating tooling so it is more clear on which tool goes with which vendor.

**Describe alternatives you've considered**
Manual data entry.

**Additional context**
Nothing

PR #158, opened 2024-02-20
Implemented importer for Arista's show inventory

cmwill:
#141

### DIFF
--- a/datastore/importers/nmdb-import-arista-show-inventory/Parser.cpp
+++ b/datastore/importers/nmdb-import-arista-show-inventory/Parser.cpp
@@ -343,6 +343,7 @@ Parser::Parser() : Parser::base_type(start)
     token =
         +qi::ascii::graph
     ;
+        (handleSection)
 
     ignoredLine =
         (+token > -qi::eol) | +qi::eol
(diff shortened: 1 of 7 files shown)


marshall-sg (changes requested):
There are changes needed as, at least, a couple of potential scenarios were missed.  Also, there is no unit test logic.  The parser should have a file called `Parser.test.cpp` which contains test logic, primarily, for the parser rules.

datastore/importers/nmdb-import-arista-show-inventory/Parser.cpp
```diff
@@ -0,0 +1,461 @@
...
+    ;
+
+    systemInformationStart =
+        qi::lit("System information") >>
+        qi::eol >> 
+        systemInformation
+        [(
+            pnx::bind(
+                [this](DevInfo &devInfo){
+                    devInfo.setModel(systemInformationEntry.at(0)["Model"]);
```
marshall-sg, line 188: The value for vendor should probably also be set here.

marshall-sg (changes requested):
Overall, several change requests.  They are primarily focused around easing ability to more quickly comprehend what is going on.  Thanks.

marshall-sg (changes requested):
What's here looks fine.  However, the parser still needs unit testing logic.

marshall-sg (approved):
Looks good.

PR #163, opened 2024-06-18
Show inventory bugfix

marshall-sg:
Adds parser unit testing and fixes an issue where the parser would not ignore "garbage" data around the relevant data.


marshall-sg (commented):
Initial review check.

PR #164, opened 2024-06-20
Show neighbor bugfixes

marshall-sg:
This PR resolves both logic and test coverage type issues, to include:
- Updated logic to support NXOS output format for `show ipv6 neighbors`
- Updated unit testing to cover NDP parsing logic for both Cisco IOS and NXOS
- Consolidated Cisco IOS NDP logic to handle both details and none in one call


marshall-sg (commented):
First pass review.

PR #165, opened 2024-07-09
Bugfixes cisco show ip route

marshall-sg:
This PR primarily resolves lack of format support for various Cisco `show ip route` calls.  Now,
- It can handled IPv4 and IPv6 for IOS, NXOS, and an old IOS output style.
- The IpAddress and associated parser logic supports cases where IPv4 addresses are both codified or embedded in IPv6.
- Routes are considered valid if a destination subnet and out interface are known.


marshall-sg (commented):
initial review

PR #173, opened 2024-08-30
Updates for `show cdp neighbor`

marshall-sg:
This PR resolves issues involving `show cdp neighbor`.  Primarily:
- Captures outgoing port names to be used for port-to-port connectivity identification.
- An issue where an incoming port name with a space would fail to parse.
- Adds an object to process/store port-to-port link connections between devices.  Some of the logic was already in the DB, but nothing was leveraging it.
- Bumps the libpqxx library version being compiled with/against.


marshall-sg (commented):
Review checkpoint.

marshall-sg (commented):
Next review cycle.

PR #177, opened 2024-09-13
153 improve nmdb import cisco parsing logic

cmwill:


datastore/importers/nmdb-import-cisco/Parser.test.cpp
```diff
@@ -0,0 +1,829 @@
...
+    for (const auto& test : nocdpTestDataOk) {
+      BOOST_TEST(nmdp::test(test, parserRule, blank),
+                 "Parse rule 'nocdp':" << test);
+    }
+
+    for (const auto& test : nocdpTestDataBad) {
+      BOOST_TEST(!nmdp::test(test, parserRule, blank),
+                 "Parse rule '!nocdp':" << test);
+    }
+  }
```
marshall-sg, line 95: Primarily, this should ensure `globalCdpEndabled` is not accidentally flipped to `false`.  Again, this comment generally applies throughout this file.

marshall-sg (changes requested):
Several of the changes were too restrictive and need reverted.  However, the unit test logic is primarily problematic as it is only looking at if it parsed or not, versus checking if state changed or not.  Really the point of those tests are to ensure state changes when we intend and not otherwise.

PR #182, opened 2024-11-13
modify parser to accept correct values when parsing age values

raulcruise:
add tilde character to characters accepted for age in the age >> secure >> ntfy format


raulcruise:
I'm wondering if tilde could also appear in the age value in the other format(learn >> age)

marshall-sg:
> I'm wondering if tilde could also appear in the age value in the other format(learn >> age)

Not sure, please research it and go from there.

marshall-sg (changes requested):
The current change looks fine and seems to grab things.  However, the unit testing logic needs to be updated to account for and test the change.

